PR #7, opened 2021-02-24
Replace dependency on GSL with regression tests and set MSRV to 1.42

### LINKED ISSUE #4: Friction compiling tests on Ubuntu
Not sure if this can be made simpler for Ubuntu users by switching to lower versions, just wanted to share my installation experience.

So there is no GSL-3.0 version for Ubuntu
https://github.com/Luthaf/rascaline/blob/ca9256756c92e0e150fdbe7477be0fe1852aaf1e/rascaline/Cargo.toml#L33
since the newest stable version is 2.6 according to https://www.gnu.org/software/gsl/
For a temporary fix of below error I switched it to 2.0.1
```
error: aborting due to 2070 previous errors

For more information about this error, try `rustc --explain E0658`.
error: could not compile `GSL`.

Caused by:
  process didn't exit successfully: `rustc --crate-name rgsl /home/alexgo/.cargo/registry/src/github.com-1ecc6299db9ec823/GSL-3.0.0/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type dylib --crate-type rlib --emit=dep-info,link -C prefer-dynamic -Cembed-bitcode=no -C debuginfo=2 -C metadata=f1559776a8bfee96 -C extra-filename=-f1559776a8bfee96 --out-dir /home/alexgo/lib/rascaline/target/debug/deps -L dependency=/home/alexgo/lib/rascaline/target/debug/deps --extern gsl_sys=/home/alexgo/lib/rascaline/target/debug/deps/libgsl_sys-8f5b7f66dff3843b.so --extern gsl_sys=/home/alexgo/lib/rascaline/target/debug/deps/libgsl_sys-8f5b7f66dff3843b.rlib --extern paste=/home/alexgo/lib/rascaline/target/debug/deps/libpaste-07ae39ea28b6635f.so --cap-lints allow` (exit code: 1)
warning: build failed, waiting for other jobs to finish...
```

I had to update my cmake version to a version above the currently one offered by standard apt repositories
```CMake 3.11 or higher is required.  You are running version 3.10.2```
(Helpful installation instruction https://askubuntu.com/a/865294)

Same for rust version
```
  your Rust installation is too old (you have version cargo 1.46.0

  ), at least 1.48.0 is required
```
Can be simply installed https://doc.rust-lang.org/cargo/getting-started/installation.html
but cmake apparently cannot find cargo
```
-- CARGO_STATUS No such file or directory CARGO_VERSION      CMAKE_SYSTEM_PROGRAM_PATH 
CMake Error at ../../../CMakeLists.txt:47 (message):
  could not run cargo, please make sure the Rust compiler is installed (see
  https://www.rust-lang.org/tools/install)
```
I linked a the binary in the home directory to /usr/local/bin but still nothing.
I am not sure how cmake finds programs usually, there are a lot of steps which can be checked https://cmake.org/cmake/help/latest/command/find_program.html
Here I gave up and just hide the `rascaline-c-api/tests/c-api.rs` file and it works :)

Luthaf:
This should fix #4


Luthaf:
As a nice side effect, this brings the time to run tests from 30s to below 10s, in debug mode 🎉

PR #11, opened 2021-03-09
Allow passing log events from rust to C/Python

agoscinski:
Only a first draft which runs and passes test. I dont think that it it works as intended. Needs to be tested somehow. If I use 
```rust
system.callback()("Hello")
```
nothing is printed in the logs.

In addition:
- need merge
- remove changes in CMakesLists.txt, pyproject.toml, rascaline-c-api/tests/c-api.rs and Cargo.toml


rascaline-c-api/src/lib.rs
```diff
@@ -8,7 +8,56 @@ mod utils;
...
+
+pub type RascalLoggingCallback = Option<unsafe extern fn(level: i32, message: *const std::os::raw::c_char)>;
+
+
+static mut GLOBAL_CALLBACK: RascalLoggingCallback = None;
+struct RascalLogger;
+
+#[no_mangle]
+pub unsafe extern fn rascal_set_logging_callback(callback: RascalLoggingCallback) {
+    GLOBAL_CALLBACK = callback;
```
Luthaf, line 27: This needs to go through a Mutex to prevent data races if two threads want to set callback at the same time

rascaline-c-api/src/lib.rs
```diff
@@ -8,7 +8,56 @@ mod utils;
...
+    log::set_boxed_logger(Box::new(RascalLogger)).expect("Setting rascal logger faild.")
```
Luthaf, line 28: This can not use expect because we want to be able to call `rascal_set_logging_callback` multiple times. Instead you can use something like OnceCell or just warn if the result is an error

rascaline-c-api/src/lib.rs
```diff
@@ -42,22 +48,12 @@ impl log::Log for RascalLogger {
                 record.args());
             let cstr = CString::new(message).unwrap();
             unsafe {
-              GLOBAL_CALLBACK.unwrap()(record.level() as i32, cstr.as_ptr());
+                //let mut guard = GLOBAL_CALLBACK.lock().unwrap();
+                //let mut f = guard.unwrap();
+                GLOBAL_CALLBACK.lock().unwrap().unwrap()(record.level() as i32, cstr.as_ptr());
```
Luthaf, line 53: first unwrap should be an `expect("mutex was poisonned")` or similar. We want to handle the second unwrap explicitly: if `GLOBAL_CALLBACK` is None, that means that the function pointer is NULL, and we don't want to call it. Instead we will want to skip sending the message since no logger was setup.

python/rascaline/clib.py
```diff
@@ -15,8 +16,36 @@ def __call__(self):
...
             setup_functions(self._cache)
+            _set_default_logging_callback()
         return self._cache
 
+def _set_default_logging_callback():
+    '''
+    Default logging function. For now print is used.
+    '''
+    def default_print(level, string):
+        print(string)
```
Luthaf, line 27: Instead of `print`, I think the default should use the [logging](https://docs.python.org/3/library/logging.html) module for better integration with the ecosystem

python/rascaline/clib.py
```diff
@@ -15,8 +16,36 @@ def __call__(self):
...
+
+    def callback(level, message):
+        try:
+            function(str(level), message.decode("utf8"))
+        except Exception as e:
+            message = "exception raised in logging callback: {}".format(e)
+            warnings.warn(message, ResourceWarning)
+
+    global _CURRENT_CALLBACK
+    _CURRENT_CALLBACK = ctypes.CFUNCTYPE(None, ctypes.c_int, ctypes.c_char_p)(callback)
```
Luthaf, line 45: Instead of rewriting the function type, could this import and use `RascalLoggingCallback` from _rascaline.py?

python/tests/calculators.py
```diff
@@ -5,10 +5,35 @@
...
 
 
+
 class TestDummyCalculator(unittest.TestCase):
+    def test_callback(self):
+        recorded_levels = []
+        recorded_messages = []
+        def test_callback(level, message):
+            recorded_levels.append(level)
+            recorded_messages.append(message)
```
Luthaf, line 21: This might be easier to test by combining level & messages with something like
```suggestion
        recorded_log_events = []
        def test_callback(level, message):
            recorded_log_events.append((level, message))
```

And then check that we get `(log_level_info, targeted_message)` in the recorded events

python/tests/calculators.py
```diff
@@ -5,10 +5,35 @@
...
+        test_callback_ctype = ctypes.CFUNCTYPE(None, ctypes.c_int, ctypes.c_char_p)(test_callback)
+        _get_library().rascal_set_logging_callback(test_callback_ctype)
+
+        system = TestSystem()
+        calculator = DummyCalculator(cutoff=3.2, delta=2, name="", gradients=True)
+        descriptor = calculator.compute(system)
+
+        print("recorded_messages", recorded_messages)
+        targeted_message = b"INFO:rascaline::calculators::dummy_calculator -- Computation of DummyCalculator is invoked."
```
Luthaf, line 30: This should not have to be a byte string, but a standard python string instead

rascaline-c-api/rascaline.h
```diff
@@ -52,6 +52,8 @@ typedef struct rascal_calculator_t rascal_calculator_t;
  */
 typedef struct rascal_descriptor_t rascal_descriptor_t;
 
+typedef void (*RascalLoggingCallback)(int32_t level, const char *message);
```
Luthaf, line 55: I would rather use C naming convention for this C type, i.e. 
```suggestion
typedef void (*rascal_logging_callback_t)(int32_t level, const char *message);
```

rascaline/src/calculators/dummy_calculator.rs
```diff
@@ -75,6 +77,8 @@ impl CalculatorBase for DummyCalculator {
 
     #[allow(clippy::clippy::cast_precision_loss)]
     fn compute(&mut self, systems: &mut [&mut dyn System], descriptor: &mut Descriptor) {
+        // Thes message of this log is tested, therefore if changed here, it has to be changed also in the tests
+        info!{"Computation of DummyCalculator is invoked."}
```
Luthaf, line 81:
```suggestion
        info!("this is an info message used for testing purposes");
        warn!("this is a warning message used for testing purposes");
```

This way we can also check passing the logging level!

rascaline-c-api/rascaline.h
agoscinski: I think if you rename it here, you have to rename it also in the rust file, and then you get a warning

rascaline/src/calculators/dummy_calculator.rs
agoscinski: The current message was chosen to make it more informative , so it can be also considered as info message for the other tests. But even then it was a bit annoying to see it printed out in every test. Therefore I have set the default log level to warnings so one only sees the message when changing the default (I do not checked warnings at the moment).

I think in the long run a dedicated function on the rust side would be the best option. For a temporary solution I can change it the suggestion. I guess if we also want to check the warnings then there is no way of creating an informative message, so what you suggest is currently the best option.

python/tests/calculators.py
agoscinski: changed

python/tests/calculators.py
agoscinski: changed

python/rascaline/clib.py
agoscinski: Yes this was the plan, I thought this would be the next PR, but actually its very simple so I changed to logging

python/rascaline/clib.py
agoscinski: hm... this works, I guess this is okay like this
```
    _CURRENT_CALLBACK = _get_library().rascal_set_logging_callback.argtypes[0](callback_wrapper)
```

rascaline/src/calculators/dummy_calculator.rs
agoscinski: changed

rascaline-c-api/rascaline.h
Luthaf: Yes, but that's a warning we can ignore 😄. You can use something like 

```rust
#[allow(non_camel_case_types)]
pub type rascal_logging_callback_t = Option<unsafe extern fn(level: i32, message: *const std::os::raw::c_char)>;
```

rascaline-c-api/src/lib.rs
```diff
@@ -8,7 +8,65 @@ mod utils;
...
+// Mutex cannot use rust static
+// see https://stackoverflow.com/a/27826181
+lazy_static! {
+    static ref GLOBAL_CALLBACK: Mutex<RascalLoggingCallback> = Mutex::new(None);
+}
+
+struct RascalLogger;
+
+#[no_mangle]
+pub unsafe extern fn rascal_set_logging_callback(callback: RascalLoggingCallback, log_level: i32) {
```
Luthaf, line 31: I'm not sold on exposing the `max_log_level` to the end user directly (in the second parameter here). I think the main use case for this feature is to make sure warnings are visible by the end user; and secondary use case is to give them some info on the running calculation (i.e. `info!` logs). 

Both use cases are well served with simpler code that do not need to deal with the max log level.

python/rascaline/log_utils.py
```diff
@@ -0,0 +1,27 @@
...
+    elif log_level == RUST_LOG_LEVEL_WARN:
+        logging.warning(message)
+    elif log_level == RUST_LOG_LEVEL_INFO:
+        logging.info(message)
+    elif log_level == RUST_LOG_LEVEL_DEBUG:
+        logging.debug(message)
+    elif log_level == RUST_LOG_LEVEL_TRACE:
+        print(message)
+    else:
+        raise ValueError('Log level ' + str(log_level) + ' is not supported.')
```
Luthaf, line 27:
```suggestion
        raise ValueError(f'Log level {log_level} is not supported.')
```

python/rascaline/log_utils.py
```diff
@@ -0,0 +1,27 @@
...
+RUST_LOG_LEVEL_OFF = 0
+RUST_LOG_LEVEL_ERROR = 1
+RUST_LOG_LEVEL_WARN = 2
+RUST_LOG_LEVEL_INFO = 3
+RUST_LOG_LEVEL_DEBUG = 4
+RUST_LOG_LEVEL_TRACE = 5
+
+DEFAULT_LOG_LEVEL = RUST_LOG_LEVEL_WARN
+
+logging.basicConfig(filename='rascaline.log', level=DEFAULT_LOG_LEVEL)
```
Luthaf, line 13: I don't think we want to write to files by default. If the end user want to do this, they can; but I would not pollute the filesystem if possible.

python/rascaline/clib.py
Luthaf: oh right. One simpler way to do so is to add 

```
rascal_logging_callback_t = ctypes.CFUNCTYPE(None, ctypes.c_int, ctypes.c_char_p)
```

around hand-written type definition https://github.com/Luthaf/rascaline/blob/cc7a5b98357825e45da8ac9b8e89453801f622f0/python/rascaline/_rascaline.py#L14-L19, 

and then use `rascal_logging_callback_t` in `rascal_set_logging_callback.argtypes`. This way, you can also `from _rascaline import rascal_logging_callback_t` where needed

python/tests/calculators.py
Luthaf: I don't see it?

rascaline/src/calculators/dummy_calculator.rs
Luthaf: > I think in the long run a dedicated function on the rust side would be the best option. For a temporary solution I can change it the suggestion. I guess if we also want to check the warnings then there is no way of creating an informative message, so what you suggest is currently the best option.

I had another idea to reduce the noise: you can create a DummyCalculator with the `string` (hyper)-parameters set to `log-test`, and then check the parameter before emitting the log messages.

Luthaf (commented):
I'm not convinced we need to expose the max log level here. For which use case did you added this?

rascaline-c-api/src/lib.rs
agoscinski: In any case the end user can just filter the logs in his/her logging utility. So setting the max log level to the maximum level and let the user decide what to filter out, this was the first solution I implemented. But then I was worried that if one has a huge trace log at some point, it could cause performance issues because it has to be forwarded to the logger of the end user before it is decided to not log it. So I thought that the end user should have the possibility to filter out beforehand on the rust side.

agoscinski: In the calculator was a comment that crossing the FFI boundary often can cause performance issues, and forwarding a lot of messages seemed to me like a possible problem in future. We could also enforce that the verbose messages are within a `debug_assertion` block, but then I don't know how well this works with other rust libraries. So assume we import a crate which uses !debug not in `debug_assertion` blocks, then we again have the original problem. I am not sure if this are really concerns which could happen in practice.

Luthaf: About performance, I don't forsee us sending 10 000 log messages per structures, but there can easily be this many neighbors. So performance might not be an issue for logging =)

About having a lot of debug/trace messages coming from other crates, we could set the max log level to `info` when registering the boxed logger. 

We can always revisit this decision if logging ends up being a performance issue down the line. But in general I find that not trying to optimize until we need it result in cleaner and easier to understand code.

agoscinski: But for debugging purpose you want to be able to set it to a higher log levels. Also  there might be case where you want a log message on a per structure level for debugging, of course while you are debugging you would only run it on one structure, but then you need to control the log level.
What do you think about: Set it by default to info, as you said, with an additional `debug_assertion` which sets it to debug level if compiled in debug mode?

Luthaf: Using `cfg=debug_assertions` for this sounds good! 

The main tool I use for temporary debugging is the [dbg!](https://doc.rust-lang.org/std/macro.dbg.html) macro, which sends output to the standard error anyway, and is not passed to Python. With this PR merged, I can see myself adding temporary `info!` or `debug!` calls instead =)

python/rascaline/clib.py
agoscinski: Not sure if this is simpler, because you need to change the the generation script in a transparent way. In the end it will be more code, might be more understandable but I am not convinced enough to go through it.

rascaline/src/calculators/dummy_calculator.rs
agoscinski: yes seems just like a lot of work for one additional test

python/tests/calculators.py
agoscinski: still open, forgot this

Luthaf: done

rascaline/src/calculators/dummy_calculator.rs
Luthaf: done

Luthaf:
I rebased this branch and made a bunch of small changes. I tried to keep every commit focused and small, so that you can go through them and we can discuss each change.

rascaline-c-api/tests/c-api/logging.cpp
```diff
@@ -0,0 +1,76 @@
...
+    CHECK_SUCCESS(rascal_calculator_compute(
+        calculator, descriptor, &system, 1, options
+    ));
+
+    rascal_calculator_free(calculator);
+    rascal_descriptor_free(descriptor);
+}
+
+TEST_CASE("Logging") {
+    auto record_log_events = [](int level, const char* message) {
```
agoscinski, line 33: weird when I tried a lambda function, it did not work because of type problems

rascaline-c-api/src/logging.rs
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+    }
+
+    fn flush(&self) {}
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
agoscinski, line 95: what does this do?

rascaline/src/calculators/dummy_calculator.rs
```diff
@@ -72,6 +74,12 @@ impl CalculatorBase for DummyCalculator {
     #[allow(clippy::clippy::cast_precision_loss)]
     #[time_graph::instrument(name = "DummyCalculator::compute")]
     fn compute(&mut self, systems: &mut [Box<dyn System>], descriptor: &mut Descriptor) -> Result<(), Error> {
+        if self.name.contains("log-test-info:") {
```
agoscinski, line 77: seems like a nice solution

rascaline-c-api/src/logging.rs
```diff
@@ -0,0 +1,105 @@
...
+/// function will be called on all log events. If a logging callback was already
+/// set, it is replaced by the new one.
+#[no_mangle]
+pub unsafe extern fn rascal_set_logging_callback(callback: rascal_logging_callback_t) -> rascal_status_t {
+    catch_unwind(|| {
+        *GLOBAL_CALLBACK.lock().expect("mutex was poisoned") = callback;
+        // we allow multiple sets of logger, therefore the result will be ignored
+        let _ = log::set_boxed_logger(Box::new(RascalLogger));
+
+        if cfg!(debug_assertions) {
```
agoscinski, line 61: also much nicer solution than before

rascaline-c-api/tests/c-api/logging.cpp
Luthaf: Capturing lambda do not work, so the recorded events still need to be a global static variable. Non-capturing lambdas can be used as function pointers.

rascaline-c-api/src/logging.rs
Luthaf: Checking that the values exported above actually match the values from `log` (in case the values from log change at some point in the future, or we want to use a separate logging facade). Should I add a small comment explaining this?

rascaline-c-api/src/logging.rs
Luthaf: yeah, that's a trick which is very rarely useful, but nice when you can use it!

Luthaf (approved):
This is good to go! Let me know if you want to make other changes, otherwise I'll squash & merge tomorrow.

rascaline-c-api/src/logging.rs
Luthaf: Oh right, I though you asked about the test module =). `use super::*;` is equivalent to python `import * from ..`. Like Python, using wildcard imports is usually discouraged, except in this exact situation: importing everything from the parent module to write corresponding tests. 

Without this, I could have written `use super::RASCAL_LOG_LEVEL_DEBUG` (similar to `from .. import RASCAL_LOG_LEVEL_DEBUG)`, or use the full access path from the crate root: `use crate::logging::RASCAL_LOG_LEVEL_DEBUG;` (similar to `from rascaline.logging import RASCAL_LOG_LEVEL_DEBUG`).

PR #38, opened 2021-05-26
Expose Indexes::positions to C/C++/Python

Luthaf:
This is required to efficiently deal with gradients/values and might help writing new calculators in Python


Luthaf:
This was not needed for what I wanted to do, so I'll close the PR but keep the code around in case it is needed later

PR #40, opened 2021-06-02
Separate debug & release build in CI

Luthaf:
This should make CI run faster overall.


Luthaf:
CI went from 26 min per job to 10-15 min per job (in debug-release mode respectively)

PR #49, opened 2021-07-19
Also enable parallelism over structures in spherical expansion

Luthaf:


Luthaf:
This is actually slower than the implementation on master. I though is was faster yesterday but it was only computing the spherical expansion for 1 system out of the 10 in the benchmarks ...

PR #68, opened 2022-02-09
Allow to provide an empty set of selected sample/features

Luthaf:
The main use case is when we have a MD domain with no atoms inside, only "ghosts" atoms from neighboring domains

### DIFF
--- a/rascaline-c-api/src/calculator.rs
+++ b/rascaline-c-api/src/calculator.rs
@@ -180,18 +180,18 @@ pub struct rascal_calculation_options_t {
     /// faster than having to cross the FFI boundary too often.
     use_native_system: bool,
     /// List of samples on which to run the calculation. You can set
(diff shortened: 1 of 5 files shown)


rascaline-c-api/src/calculator.rs
```diff
@@ -180,18 +180,18 @@ pub struct rascal_calculation_options_t {
...
     /// List of features on which to run the calculation. You can set
-    /// `selected_features.values` to `NULL` to run the calculation on all
+    /// `selected_features.names` to `NULL` to run the calculation on all
     /// features.
     selected_features: rascal_indexes_t,
 }
 
 fn selected_indexes(selected: &rascal_indexes_t) -> Result<SelectedIndexes, Error> {
-    if selected.values.is_null() {
+    if selected.names.is_null() {
```
Luthaf, line 194: this is the only code change, everything else is documentation & tests 😁

PR #72, opened 2022-04-20
Fixed little typo in comment for GTO width

PicoCentauri:
Change the comment according to the expression below eq. 10 of [Musil et. al](https://aip.scitation.org/doi/10.1063/5.0044689). The actual calculation in the code is correct.


Luthaf (approved):
Thanks!

PR #75, opened 2022-05-02
Added build and install test

### LINKED ISSUE #5: Add 'install' tests on CI
Basically simple tests that check that both `pip install .`; and that `mkdir build && cd build && cmake .. && make install` work.

PicoCentauri:
Closes #5


tox.ini
```diff
@@ -1,67 +1,68 @@
...
 commands =
     python setup.py --quiet install
     discover -p "*.py" -s python/tests
 
 [testenv:min-deps]
-skip_install = True
-passenv =
-    RASCALINE_BUILD_TYPE
-    RUSTUP_HOME
-    CARGO*
```
Luthaf, line 36: `RASCALINE_BUILD_TYPE` is required to be able to run the python tests in either debug or release mode. 

This repository has been initially built to have cargo as the main test runner for everything, and in particular running `cargo test --all` in the root will run [tox and all the python tests](https://github.com/Luthaf/rascaline/blob/master/rascaline-c-api/tests/python-api.rs). `RASCALINE_BUILD_TYPE` is used to pass the information of "build in debug/release mode" from cargo back to cargo through tox => setup.py => cmake => cargo. This is a bit hacky and we could have a discussion on the best way to organize all test & pick a main test runner, but I think this needs to stay around for now.

---

The two other variables are requires in case someone has a non-standard setup for rustup/cargo installation.

tox.ini
```diff
@@ -1,67 +1,68 @@
...
+    build
     all-deps
     min-deps
     examples
-
-
-[testenv:all-deps]
-# skip install to avoid rebuilding the world, do the install with setup.py
-# instead of pip below
-skip_install = True
```
Luthaf, line 15: How do build time look without this? This was an optimization before pip used in-tree build to be able to re-use cached object files in `target`. You can check this by comparing the time it takes to do `pip install .` after `rm -rf target/` (no cache) and after `rm -rf target && cargo build --release`. If the second one is much smaller, then we don't need this `skip_install` anymore.

tox.ini
```diff
@@ -1,67 +1,68 @@
...
+    make
+    mkdir
+deps =
+    setuptools
+    wheel
+    twine
+    check-manifest
+skip_install = true
+commands =
+    mkdir -p build && cd build && cmake .. && make install
```
Luthaf, line 58: In addition to what was in the issue, it would be a good idea to try to build a small example cmake project using the installed library, making sure the installed files are actually usable.

This might require some more filesystem management, with a `CMAKE_INSTALL_PREFIX=<somepath/somewhere>`, and pointing cmake there when building the example project.

---

Also, instead of using make which is not available everywhere I would use `cmake --build . --target install`. Depending on how familiar you are with cmake, I'm happy to help you do something like this or do it myself.

tox.ini
Luthaf: oh, just realized you moved the setting on top of the tox.ini ...

tox.ini
PicoCentauri: I thought one could it pass globally. But this doesn't seem the case...

tox.ini
PicoCentauri: I am not a expert in cmake building but I can specify a path.

Luthaf:
I just rebased and updated this branch with tests checking that we can build, install and use the code from a cmake project.

tox.ini
```diff
@@ -45,41 +60,43 @@ commands =
...
 commands =
-    flake8 {posargs:python setup.py}
-    black --check --diff .
-    isort --check-only --diff python
+    # check building sdist and wheels from a checkout
+    python setup.py sdist
+    python setup.py bdist_wheel
+    twine check dist/*.tar.gz
+    twine check dist/*.whl
+
```
Luthaf, line 89: @PicoCentauri I had to remove check-manifest since it was failing to run inside a container (see https://github.com/Luthaf/rascaline/runs/6940873009?check_suite_focus=true, which had `git config --global --add safe.directory /__w/rascaline/rascaline`).

I think it is fine since we are testing that the sdist can be used to build a wheel just below

PicoCentauri:
Yes should be fine.

>

PR #76, opened 2022-05-06
Remove double normalization from GTO radial basis

Luthaf:


Luthaf:
All tests are expected to fail since I have to also change the reference data in regression tests.

PR #39, opened 2021-05-28
Multithreaded calculation of spherical expansion and power spectrum

Luthaf:
The strategy used is to parallelize over pairs for spherical expansion, and accumulate samples in parrallel for the power spectrum. Most of the parallelism comes from [rayon](https://docs.rs/rayon/1.5.0/rayon/) operations.

# Benchmark comparisons versus master

### RAYON_NUM_THREADS=1

There is a cost to the parallel infrastructure, with benchmarks with low n_max/l_max being up to 4 times slower, while benchmarks with high n_max/l_max are up to 30% faster!

<details>

```
group                                                                                      1-thread                 master
-----                                                                                      --------                 ------
Spherical expansion (per atom) with gradients/Bulk Silicon/n_max = 14, l_max = 1           1.48     96.2±1.60µs     1.00     64.8±0.78µs
Spherical expansion (per atom) with gradients/Bulk Silicon/n_max = 14, l_max = 15          1.00  1333.8±15.86µs     1.34  1781.9±14.38µs
Spherical expansion (per atom) with gradients/Bulk Silicon/n_max = 14, l_max = 7           1.00    378.6±0.51µs     1.14   431.3±11.70µs
Spherical expansion (per atom) with gradients/Bulk Silicon/n_max = 2, l_max = 1            4.28     85.0±0.38µs     1.00     19.9±0.23µs
Spherical expansion (per atom) with gradients/Bulk Silicon/n_max = 2, l_max = 15           1.00    261.6±1.00µs     1.08    283.7±1.54µs
Spherical expansion (per atom) with gradients/Bulk Silicon/n_max = 2, l_max = 7            1.48    110.7±0.56µs     1.00     75.0±0.76µs
Spherical expansion (per atom) with gradients/Bulk Silicon/n_max = 8, l_max = 1            2.37    100.7±0.39µs     1.00     42.5±0.31µs
Spherical expansion (per atom) with gradients/Bulk Silicon/n_max = 8, l_max = 15           1.00   919.5±27.80µs     1.16   1066.1±8.07µs
Spherical expansion (per atom) with gradients/Bulk Silicon/n_max = 8, l_max = 7            1.00    265.7±2.42µs     1.06    281.6±0.67µs
Spherical expansion (per atom) with gradients/Molecular crystals/n_max = 14, l_max = 1     1.40    175.3±4.41µs     1.00    125.6±0.74µs
[...]

PR #78, opened 2022-05-27
The PR that changed everything

Luthaf:
This removes the internal Descriptor type and replace it by equistore as a native output.

Still to be done:
- [x] Update the tutorial "how to write a calculator"
- [ ] Decide what to do with C++ API: remove it or add a C++ API to equistore & update the rascaline C++ API.
- [x] Update the regression test data

I also had to change the parallelisation strategy, benchmarks incoming!


Luthaf:
Here are some strong scaling benchmarks:

molecular_crystals-strong-scaling-single

To compare versus an old version of the same benchmarks (from #39):
old-scaling

Luthaf:
> Decide what to do with C++ API: remove it or add a C++ API to equistore & update the rascaline C++ API.

I'll update the C++ API later, as far as I know nobody is using it for now.

PR #81, opened 2022-06-23
Cell gradients

Luthaf:
TODO: 

- [x] add more tests? There are only some finite differences tests, we may want to have at least some regression tests as well.


DavideTisi:
Hey,
 I just started looking at it. Everything seems fine. One general question.
I see that the choice of computing or not a gradient is terminated by two distinct input parameters in the CalculationOptions: `position_gradient` and `cell_gradient`. Why not instead, we could not use a single parameter like a "list" of str like `['positions', 'cell']` and if the string is present allow the computation or not. If in the future new gradients are required we could allow a new option e so on.

I do not know, just something that comes to mind, what do you think?
I even do not know if it easily doable in rust

Luthaf:
So using a list of parameters when selecting the gradients would work fine. It would be slightly harder to pass data through the C API, but still manageable.

After playing a bit with the two options, I think I like the version with a single `gradients=["positions", "cell"]` parameter instead of multiple options.

What do others think? @ceriottm @PicoCentauri @agoscinski @kvhuguenin

The two options are multiple boolean parameters:
```py
# both gradients
descriptor = calculator.compute(frames, positions_gradient=True, cell_gradient=True)

# single gradient
descriptor = calculator.compute(frames, positions_gradient=True)

# selected samples/properties
descriptor = calculator.compute(
    frames, 
    positions_gradient=True,
    cell_gradient=True,
    selected_samples=samples,
    selected_properties=properties,
)
```

Or a single list of strings parameter:
```py
# both gradients
descriptor = calculator.compute(frames, gradients=["positions", "cell"])

# single gradient
descriptor = calculator.compute(frames, gradients=["positions"])

# selected samples/properties
descriptor = calculator.compute(
    frames, 
    gradients=["positions", "cell"],
    selected_samples=samples,
    selected_properties=properties,
)
```

PicoCentauri:
If there are not more than two options I would go with independent parameters (option 2). If we plan even more option a list could be a good option.

rascaline/examples/compute-soap.rs
```diff
@@ -17,7 +17,6 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         "max_angular": 4,
         "atomic_gaussian_width": 0.3,
         "center_atom_weight": 1.0,
-        "gradients": true,
```
DavideTisi, line 20: Now, this example does not compute the gradients. It never used them but in that sense it is different. Probably leaving the computations in the example (even without using the gradients) would show a user how to do it.

rascaline/examples/profiling.rs
```diff
@@ -50,7 +49,12 @@ fn compute_soap(path: &str) -> Result<TensorMap, Box<dyn std::error::Error>> {
 
     let mut descriptor = time_graph::spanned!("Full calculation", {
         let mut calculator = Calculator::new("soap_power_spectrum", parameters.to_owned())?;
-        calculator.compute(&mut systems, Default::default())?
+
+        let options = CalculationOptions {
+            gradients: &["positions"],
+            ..Default::default()
+        };
+        calculator.compute(&mut systems, options)?
```
DavideTisi, line 57: For example here the computation of gradients is correctly set.

ceriottm:
Did you already check that numerical derivatives match the analytical ones?

DavideTisi:
Not really sure what you meant, but there is a test that checks the computed gradients against a "finite difference" gradient

PR #83, opened 2022-06-28
Fix doc building and make sure sphinx warnings are treated as errors

Luthaf:


Luthaf:
The documentation build was missing equistore.

I added `-W` to the sphinx flags to make sure this does not go unnoticed next time.

PR #88, opened 2022-07-07
Implemenet k-vector computation

PicoCentauri:
Added computation of k vectors for LODE calculations. I will add the tests but if @Luthaf or @kvhuguenin already have ideas please let me know.

### DIFF
--- a/rascaline/src/math/kvectors.rs
+++ b/rascaline/src/math/kvectors.rs
@@ -14,7 +14,8 @@ pub struct KVector {
 }
 
 pub fn compute_kvectors(cell: &Matrix3, kcutoff: f64) -> Vec<KVector> {
-    // Generate k-vectors up to a certain cutoff for a given cell
+    // Generate k-vectors up to a certain cutoff (in reciprocal space units) 
+    // for a given cell.
     let reciprocal_cell = 2.0 * std::f64::consts::PI * cell.transposed().inverse();
 
     let cutoff_squared = kcutoff * kcutoff;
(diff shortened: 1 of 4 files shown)


rascaline/src/math/kvectors.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// 3 component k-vector
+    pub vector: Vector3D,
+    /// length of the k-vector
+    pub norm: f64,
+}
+
+pub fn compute_kvectors(cell: &Matrix3, kcutoff: f64) -> Vec<KVector> {
+    // Generate k-vectors up to a certain cutoff (in reciprocal space units) 
+    // for a given cell.
+    let reciprocal_cell = 2.0 * std::f64::consts::PI * cell.transposed().inverse();
```
Luthaf, line 19: You should add a sanity check that the cell is defined, something like 

```suggestion
/// Generate k-vectors up to a certain cutoff (in reciprocal space units) 
/// for a given cell.
pub fn compute_kvectors(cell: &Matrix3, kcutoff: f64) -> Vec<KVector> {
    debug_assert_ne!(cell, Matrix3::zero(), "invalid cell matrix");

    let reciprocal_cell = 2.0 * std::f64::consts::PI * cell.transposed().inverse();
```

Or take a `UnitCell` as input and check the shape here!

PicoCentauri: I like the approach with the matrix, because this is really general. Thanks for the sanity checks!

Luthaf:
Looks good overall, could you rebase and squash before we merge? (Or I'll use a squash merge on github, as you prefer!)

PicoCentauri:
Nice, we can squash merge here.

Luthaf:
Sorry, I missed your message here

PicoCentauri:
No worries thanks for the merge.

PR #95, opened 2022-09-06
Refactor the SOAP radial integral

Luthaf:
This was prompted by realizing that the previous implementation of the 1F1 function did not support negative arguments. Even with a change similar to https://github.com/lab-cosmo/librascal/commit/481385986b2169c2c1311625e83c3288a9b40acc, the gradients of the 1F1 function for negative arguments are very wrong.

The first commit in this PR thus replaces our custom 1F1 implementation with the one from scipy, translated from Python/Fortran.

The resulting code is ~100x slower than the previous implementation, but should be suitable for implementing the LODE spherical integral. To counteract the slowdown, this PR also makes using splines for the radial integral the default, with a possible opt-out parameter.


PicoCentauri:
Really good. PRs that have less code in the end are always nice. Good also that also the extra python file for tests is gone!

PicoCentauri (approved):

Luthaf:
Looks like this is working for LODE as well, so I'll merge it!

PR #97, opened 2022-09-08
Implement projection of k-vectors to nlm

Luthaf:
This computes the radial integral and spherical harmonics for the k-vectors. The resulting values agree with pylode [`k_dep_factor_reordered`](https://github.com/ceriottm/lode/blob/fd2b88e3c8110c286bdb9a08908000c3dc0b4b58/src/pylode/lib/projection_coeffs.py#L346-L351) with `G_k` set to 1.0.


PicoCentauri:
LGTM!

PR #98, opened 2022-09-09
Full LODE implementation

Luthaf:
This finishes the LODE code and adds the gradients w.r.t. positions.

Still TODO before we can merge to master:
- [x] Add k=0/l=0 term for p=0 / p>3
- [ ] Add `center_atom_weight` parameter as in SOAP.
- [ ] port high p=1 level pyLODE tests (Madelung constants, Random Coulomb Structures, Super cells)
- [x] tests for p=0 against SOAP implementation
- [ ] add regression tests for `compute_k0_contributions`, `compute_density_fourrier`, `project_k_to_nlm`, `compute_structure_factors`
- [ ] enable parallelism (likely over systems like for the SOAP spherical expansion)
- [x] fix all TODO in the code
- [x] Docs


PicoCentauri:
Thanks for this intensive work @Luthaf. I will go through the code and do some initial review. I can also do these check resulting and adding the highlevel tests.

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -92,48 +78,63 @@ pub struct LodeSphericalExpansion {
...
+    let mut real_per_center = all_species.iter().copied()
+        .map(|s| (s, Array2::from_elem((n_atoms, n_k_vectors), 0.0)))
+        .collect::<BTreeMap<_, _>>();
+    let mut imag_per_center = all_species.iter().copied()
+        .map(|s| (s, Array2::from_elem((n_atoms, n_k_vectors), 0.0)))
+        .collect::<BTreeMap<_, _>>();
+
+    for i in 0..n_atoms {
+        for j in 0..n_atoms {
+            for k in 0..n_k_vectors {
```
PicoCentauri, line 120: For parallelism the plan is to use also the `.iter()` here as well instead of the explicit loop over `k`, right?

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -342,9 +390,155 @@ impl CalculatorBase for LodeSphericalExpansion {
...
+
+            let structure_factors = compute_structure_factors(
+                system.positions()?,
+                system.species()?,
+                &k_vectors
+            );
+
+            let density_fourrier = self.compute_density_fourrier(&k_vectors);
+
+            let global_factor = 4.0 * std::f64::consts::PI / cell.volume();
```
PicoCentauri, line 403: I am not happy with the name `global_factor`. Maybe we find something more descriptive. A bit better would be to call it `normalization`.

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -342,9 +390,155 @@ impl CalculatorBase for LodeSphericalExpansion {
...
+
+                                let mut value = 0.0;
+                                for ik in 0..k_vectors.len() {
+                                    // Use unsafe to remove bound checking in
+                                    // release mode with `uget` (everything is
+                                    // still bound checked in debug mode).
+                                    //
+                                    // This divides the calculation time by two
+                                    // for values.
+                                    unsafe {
```
PicoCentauri, line 450: I like unsafe casting :-)

rascaline/src/calculators/lode/spherical_expansion.rs
Luthaf: We can try making the structure factor calculation parallel as well yes! If you want to give it a go feel free to

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -353,30 +547,40 @@ impl CalculatorBase for LodeSphericalExpansion {
...
-            [[2.0, 2.0], [1.0806046117362793, 1.0806046117362793]],
-            [[1.0806046117362793, 1.0806046117362793], [2.0, 2.0]]
-        ]);
-        let expected_imag = arr3(&[
-            [[0.0, 0.0], [-1.682941969615793, -1.682941969615793]],
-            [[1.682941969615793, 1.682941969615793], [0.0, 0.0]]
-        ]);
-
-        assert_eq!(structure_factors.real, expected_real);
-        assert_eq!(structure_factors.imag, expected_imag);
```
PicoCentauri, line 380: Is there a specific reason to remove the tests here?

Luthaf: This is temporary, I should add them back in. I changed the structure factor layout a bit, removing the part which was tested here. We now store `exp(i k ri)` and `\sum_j exp(i k (rj  - ri))`, and no longer the full `exp(i k (rj  - ri))`

PicoCentauri: Okay.

rascaline/src/calculators/lode/spherical_expansion.rs
Luthaf: That works, this is just how it is also called in the SOAP spherical expansion. But we can change both!

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -22,21 +22,38 @@ use crate::math::{compute_k_vectors, KVector};
 use crate::math::{gamma_ui, gamma};
 
 
-/// TODO
+/// Parameters for LODE spherical expansion calculator.
+///
+/// The spherical expansion is at the core of representations in the LODE
+/// (long-distance equivariant) family. See [this
+/// article](https://aip.scitation.org/doi/10.1063/1.5128375) for more 
+/// information on the LODE representation, and how it is implemented in rascaline.
```
Luthaf, line 30:
```suggestion
/// information on the LODE representation.
```

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -22,21 +22,38 @@ use crate::math::{compute_k_vectors, KVector};
...
-    /// Spherical cutoff to use for atomic environments
+    /// Spherical realspace cutoff to use for atomic environments.
+    /// Note that this cutoff is only used for the projection of the density. 
+    /// In contrast to SOAP, LODE also takes atoms outside of this cutoff into 
+    /// account for the density.
     pub cutoff: f64,
+    /// Spherical reciprocal cutoff. If 0 k_cutoff
+    /// is set to 1.2 π / atomic_gaussian_width, which is reasonable value
+    /// for most systems.
+    pub k_cutoff: f64,
```
Luthaf, line 42: To make the intent clearer, we can use an Option here, and maybe make this hyper optional (see how we handle radial_scaling in SOAP)

```suggestion
    pub k_cutoff: Option<f64>,
```

rascaline/src/calculators/lode/radial_basis.rs
```diff
@@ -66,7 +66,8 @@ impl LodeRadialBasis {
                 let parameters = SplinedRIParameters {
                     max_radial: parameters.max_radial,
                     max_angular: parameters.max_angular,
-                    cutoff: parameters.cutoff,
+                    // The spline cutoff needs to be the cutoff in k-space
+                    cutoff: parameters.get_k_cutoff(),
                 };
```
PicoCentauri, line 71: @Luthaf the compiler is blaming me here that `method not found in `GtoParameters` even though we import LodeSphericalExpansionParameters. There is probably something else to import here?

Luthaf: Might need to make the function public `pub fn`

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -57,6 +57,16 @@ pub struct LodeSphericalExpansionParameters {
     pub potential_exponent: usize,
 }
 
+impl LodeSphericalExpansionParameters {
+    fn get_k_cutoff(&self) -> f64 {
+        if self.k_cutoff.is_some() {
+            return self.k_cutoff.unwrap();
+        } else {
+            return 1.2 * std::f64::consts::PI / self.atomic_gaussian_width;
+        };
```
Luthaf, line 66: Instead of using unwrap just after checking if it is some, you can use something like this:
```suggestion
        if let Some(k_cutoff) = k_cutoff {
            return k_cutoff;
        } else {
            return 1.2 * std::f64::consts::PI / self.atomic_gaussian_width;
        };
```

or something like

```suggestion
        return self.k_cutoff.unwrap_or(1.2 * std::f64::consts::PI / self.atomic_gaussian_width);
```

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -394,11 +404,7 @@ impl CalculatorBase for LodeSphericalExpansion {
     #[time_graph::instrument(name = "LodeSphericalExpansion::compute")]
     fn compute(&mut self, systems: &mut [Box<dyn System>], descriptor: &mut TensorMap) -> Result<(), Error> {
 
-        let _k_cutoff = if self.parameters.k_cutoff == 0. {
-            1.2 * std::f64::consts::PI / self.parameters.atomic_gaussian_width
-        } else {
-            self.parameters.k_cutoff
-        };
+        let k_cutoff: f64 = self.parameters.get_k_cutoff();
```
Luthaf, line 407: Any specific reason for the explicit type here?

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -194,6 +225,56 @@ impl LodeSphericalExpansion {
...
+            let factor = (4.0 * std::f64::consts::PI * smearing_squared).powf(0.75);
+
+            for k_vector in k_vectors {
+                let value = f64::exp(-0.5 * k_vector.norm * k_vector.norm * smearing_squared);
+                fourrier.push(factor * value);
+            }
+
+            // For Gaussian potentials, the Fourier transform at k = 0 is finite and contributes to the coefficients
+            // self.radial_integral.compute(0.0);
+            // let mut value = 1.0 ; // put k=0 value here
```
PicoCentauri, line 246: Sorry for asking again, but how can I access the radial integral value for k=0. 🫣

Luthaf: it is not computed currently, but you should be able to do something like `self.radial_integral.compute(0.0)` and then access `self.radial_integral.values[[l, n]]`. If it is not clear we can have a quick chat tomorrow!

PicoCentauri: Okay, yes we can have a quick chat tomorrow.

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -327,24 +394,207 @@ impl CalculatorBase for LodeSphericalExpansion {
...
+                        let sample = [system_i.into(), center_i.into()];
+                        let sample_i = match values.samples.position(&sample) {
+                            Some(s) => s,
+                            None => continue
+                        };
+
+                        // Add k = 0 for (m, l) = (0, 0)
+                        // For Gaussian potentials (potential_exponent = 0), the Fourier transform at k = 0 is
+                        // finite and contributes to the coefficients
+                        if self.parameters.potential_exponent == 1 {
```
Luthaf, line 462: this should be potential_exponent == 0
```suggestion
                        // Add k = 0 contributions for (m, l) = (0, 0)
                        if self.parameters.potential_exponent == 0 {
                            // For Gaussian potentials (potential_exponent = 0), the Fourier transform at k = 0 is
                            // finite and contributes to the coefficients
```

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -327,24 +394,207 @@ impl CalculatorBase for LodeSphericalExpansion {
...
+                            self.radial_integral.compute(0.0);
+                            for (property_i, [n]) in values.properties.iter_fixed_size().enumerate() {
+                                let n = n.usize();
+                                let value = self.radial_integral.values[[0, n]];
+                                array[[sample_i, 0, property_i]] = global_factor * prefac * value;
+                            }
+                        // For interactions decaying with potential_exponent > 4, the potential decays fast enough that the
+                        // Fourier transform of the density at k = 0 is well defined.
+                        } else if self.parameters.potential_exponent == 4 || self.parameters.potential_exponent == 5 ||
+                                self.parameters.potential_exponent == 6 {
```
Luthaf, line 477:
```suggestion
                        } else if self.parameters.potential_exponent >= 4 {
                            // For interactions decaying with potential_exponent >= 4, the potential decays fast enough that the
                            // Fourier transform of the density at k = 0 is well defined.
```

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -327,24 +394,207 @@ impl CalculatorBase for LodeSphericalExpansion {
...
+                        let values = block.values_mut();
+                        let array = values.data.as_array_mut();
+
+                        let sample = [system_i.into(), center_i.into()];
+                        let sample_i = match values.samples.position(&sample) {
+                            Some(s) => s,
+                            None => continue
+                        };
+
+                        // Add k = 0 for (m, l) = (0, 0)
```
Luthaf, line 459: I would move this whole part outside of the main `for spherical_harmonics_l in 0..=self.parameters.max_angular` loop. It looks like we are adding the contribution to all `l/m=-l` here.

Maybe a `do_k0_contributions` function?

PicoCentauri: Good idea. I added a function but now only an ownership issues has to be solved 🤯

```
error[E0502]: cannot borrow `*self` as mutable because it is also borrowed as immutable
   --> rascaline/src/calculators/lode/spherical_expansion.rs:521:42
    |
477 |                 let k_vector_to_m_n = &self.k_vector_to_m_n[spherical_harmonics_l];
    |                                        -------------------- immutable borrow occurs here
...
513 |                                             * k_vector_to_m_n.uget([m, n, ik]);
    |                                               -------------------------------- immutable borrow later used here
...
521 |                         let k0_contrib = self.compute_k0_contributions();
    |                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ mutable borrow occurs here
    ```

File: rascaline/src/calculators/lode/spherical_expansion.rs [hunk 14]
```diff
@@ -513,6 +517,13 @@ impl CalculatorBase for LodeSphericalExpansion {
                             }
                         }
 
+                        // Add k = 0 contributions for (m, l) = (0, 0)
+                        let k0_contrib = &self.compute_k0_contributions();
+                        for (_property_i, [n]) in values.properties.iter_fixed_size().enumerate() {
+                            let n = n.usize();
+                            array[[sample_i, 0, n]] += global_factor * k0_contrib[[n]];
```
Luthaf, line 524: Accessing the array should use `_property_i`, since some `n` might not have been requested by the user.

```suggestion
                            array[[sample_i, 0, _property_i]] += global_factor * k0_contrib[[n]];
```

rascaline/src/calculators/lode/spherical_expansion.rs
Luthaf: Since we only need to compute k=0 contributions for l=0, we should move this code out of the loop over `spherical_harmonics_l`. This should resolve the borrowing error at the same time =).

PicoCentauri:
@kvhuguenin and @Luthaf should we rename the former `spherical-expansion` to `soap-spherical-expansion` also in the docs? Even though SOAP is not really the expansion it might make things more clear.

ceriottm:
> @kvhuguenin and @Luthaf should we rename the former `spherical-expansion` to `soap-spherical-expansion` also in the docs? Even though SOAP is not really the expansion it might make things more clear.

you mean to distinguish from the lode version? I would rather call the lode version `spherical-expansion-kspace`

PicoCentauri:
> > @kvhuguenin and @Luthaf should we rename the former `spherical-expansion` to `soap-spherical-expansion` also in the docs? Even though SOAP is not really the expansion it might make things more clear.
> 
> you mean to distinguish from the lode version? I would rather call the lode version `spherical-expansion-kspace`

fine for me. And to be consistent:`spherical-expansion-rspace`?

Luthaf:
I started to rename everything to `soap-spherical-expansion` and `lode-spherical-expansion`, and I have a slight preference for these, but we can change it. 

> fine for me. And to be consistent: spherical-expansion-rspace?

`spherical-expansion-kspace` is nice and understandable, but `spherical-expansion-rspace` less so (`kspace` something used in a lot of places already, `rspace` or `real-space` not as much).

We could also do `spherical-expansion-soap` and `spherical-expansion-kspace`.

PicoCentauri:
I also like `spherical-expansion-kspace`.

However, this might now feel like nitpicking but everything we do is 'SOAP': One calculator is computing features in real and the other in k-space. New users might be confused and come up with an idea: `spherical-expansion-soap` is the most high level representation and includes the k-space and real-space or what ever. Therefore, I highly encourage a name scheme making clear that one is k-space and the other one is real space. 

We also might build SOAP power spectra combining real and kspace features, which should also be clear...

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -703,4 +719,35 @@ mod tests {
...
+        let mut lode_descriptor = lode_calculator.compute(&mut systems, options).unwrap();
+        let mut soap_descriptor = soap_calculator.compute(&mut systems, options).unwrap();
+
+        let keys_to_move = LabelsBuilder::new(vec!["species_center", "species_neighbor"]).finish();
+        lode_descriptor.keys_to_samples(&keys_to_move, true);
+        soap_descriptor.keys_to_samples(&keys_to_move, true);
+
+        assert_relative_eq!(lode_descriptor.blocks()[0].values().data.as_array(),
+                            soap_descriptor.blocks()[0].values().data.as_array(),
+                            max_relative=1e1);
```
Luthaf, line 751: That's not equal at all, if one can be up to 10 times the other one

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -327,24 +432,201 @@ impl CalculatorBase for LodeSphericalExpansion {
...
+
+                        let sample = [system_i.into(), center_i.into()];
+                        let sample_i = match values.samples.position(&sample) {
+                            Some(s) => s,
+                            None => continue
+                        };
+
+                        for (_property_i, [n]) in values.properties.iter_fixed_size().enumerate() {
+                            let n = n.usize();
+                            array[[sample_i, 0, _property_i]] += global_factor * k0_contrib[[n]];
```
PicoCentauri, line 487: It seems that this is not saved in the final feature array. When I compare the final features to pyLODE they only contain the 'main feature' and not the k_0 contribution. @Luthaf probably knows what's going on here.

PicoCentauri:
I am working on the Madelung Tests right now. For these tests and all tests we have to scale the center atom weight to 0. Currently this option does not exist. For GTO's there is an analytical expression which we will implement.

Luthaf:
After a discussion with @PicoCentauri we agreed to merge this in the LODE branch, and start other PR for the remaining TODO items.

PicoCentauri:
Merge this when you are ready @Luthaf and `lode` into main as well.

PR #100, opened 2022-09-15
Re-enable C++ API documentation

Luthaf:
Follow up to #78 which removed the C++ API after the move to equistore and #93 which added it back.


PicoCentauri:
Thanks for this. BTW: Is there a way to see the documentation pages of the current PR? There is a test for it but it seems that the files are not stored. We could also create a rtd page for the development docs.

Luthaf:
Unfortunately no, there is no easy way to see the docs for a PR, other than checking it out locally and building it. The test just checks that the documentation builds fine.

For now, documentation are hosted on github pages (https://luthaf.fr/rascaline/latest/index.html), and only the version corresponding to the latest master is served, but we can configure this to have more than one version online later. I find RTD a bit hard to work with when dealing with a package which is not pure Pythion, hence github pages.

Does RTD have a way to build for a specific PR? That would be neat! Even better if they can serve HTML we create in the test directly (removing all issues with dependencies when building the docs). We could build something like this on top of github actions, but it might be a lot of work.

Luthaf:
Some relevant links: https://docs.readthedocs.io/en/stable/pull-requests.html, https://github.com/community/community/discussions/7730.

Luthaf:
merging, but we can continue the discussion on making reviewing documentation changes easier!

https://dev.to/epassaro/a-rtd-like-documentation-pipeline-for-github-actions-4ci1 looks interesting as well

PR #102, opened 2022-09-15
Read the docs deploy

### LINKED ISSUE #101: Generate docs of PRs
When working with the reviewing docs it is very useful to the see the rendered results without checking out the changed locally. Read the doc allows such an option. [They can build the docs for each PR](https://docs.readthedocs.io/en/stable/pull-requests.html). The results are integrated in the Build status report of each PR.

We should add this functionality here and later also to other projects.

PicoCentauri:
Closes #101

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build-and-publish:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
         with:
@@ -43,6 +43,7 @@ jobs:
               rm -rf latest
               mv ../docs/build/html latest
           fi
+
       - name: deploy to gh-pages
         if: github.event_name == 'push'
         uses: peaceiris/actions-gh-pages@v3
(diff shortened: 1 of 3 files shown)


PicoCentauri:
I pleased the CI gods and the latest builds are [working](https://readthedocs.org/projects/rascaline/builds/18086977/).

Luthaf:
Looks nice! Is there a way to only build with rtd for pull requests (and leave the main docs on github pages)? And is there a way to post a message to the PR with a link to the documentation instead of having to know it exists?

PicoCentauri:
> Looks nice! Is there a way to only build with rtd for pull requests (and leave the main docs on github pages)? 

The main docs are stilled deployed by GH-pages. RTD is building the docs for the master branch as well. You can not disable this afaik. However, this is not so bad. We can should keep the RTD master build as a 'hidden' docs backup in case GH-pages is down or unavailable at some point.

> And is there a way to post a message to the PR with a link to the documentation instead of having to know it exists?

The results of the docs build are already integrated in the conversation and appear where the tests are. When the build is successful and you click on the `details` link you will be directed to the RDT version of the docs. If the build fails you will be redirected to the log. This is already quite handy and I think there is no need for an additional link to post.

Luthaf:
> We can should keep the RTD master build as a 'hidden' docs backup in case GH-pages is down or unavailable at some point.

I'm mostly concerned about SEO here: for search engines to direct to the right docs, you need to have a single version of them online. When you say "hidden", is there a way to make it not show up in google searches at all?

PicoCentauri:
Let me check if there is a privacy option.

Luthaf:
Ok, I got redirect to work, so that https://rascaline.readthedocs.io/en/latest/ redirects to the github pages, while the PR documentations stays at RTD.

PR #103, opened 2022-09-26
Provide pre-computed splines for the radial integrals

frostedoyster:
It works, but it needs some more work


Luthaf:
This code is not following the linter conventions (https://github.com/Luthaf/rascaline/actions/runs/3127975477/jobs/5075230898#step:11:219), you can run `tox -e format` to automatically reformat your code, and `tox -e lint` to run the linter locally.

Luthaf:
This is also missing a couple of things that should always be part of a pull request:

- [ ] user-facing documentation for the new feature
- [ ] developer documentation (documenting all the new functions)
- [ ] unit and/or regression tests

I can help with any of these, let me know if you have questions.

ceriottm:
One comment - I think it's important, to avoid confusion, that the radial integrals are defined as part of the hypers, in a way that they cannot be given together with specifications of say the basis and the Gaussian smearing. Also, we need to decide how to make the nmax(l) cutoff: we could use the radial integrals to define the expansion in full, or we could allow to define cutoffs manually. I see ups and downs for both.

Luthaf:
What do you mean by `nmax(l) cutoff`? Picking a different `n_max` value for each `l`?

ceriottm:
yes, this is an integral part of the LE basis - also, there should be
l-dependent radial integrals.

Luthaf:
Closing in favor of #140

PR #105, opened 2022-09-28
SOAP cell gradients are now actual cell gradients

Luthaf:
We previously computed gradients w.r.t. deformations


PicoCentauri (approved):

ceriottm:
Docs for this will have to be *very* clear. Many people will expect virial-like derivatives. I think doing the "natural autograd" derivative is better, but we should also explain how one converts these into epsilon derivatives.

Luthaf:
@ceriottm what do you think of the new docs?

ceriottm:
Can we add latex? I think it'd be great to write explicitly that this computed \partial <q|A> /\partial \mathbf{h} and that if one needs to recover the derivative wrt the strain epsilon (which is usually used to define the stress) they can do that as  -\partial  <q|A>/\partial\epsilon = -\partial <q|A> /\partial \mathbf{h} \cdot \mathbf{h}.

PicoCentauri:
I think it should work since we are using sphinx in the end. BTW github also [supports math now](https://github.blog/2022-05-19-math-support-in-markdown/):

$$ {\langle q \vert A \rangle}/{\partial \mathbf{h} }$$

and 

$$ -\frac{\partial \langle q \vert A \rangle}{\partial\epsilon} = -\frac{\partial \langle q \vert A \rangle}{\partial \mathbf{h} \cdot \mathbf{h}} $$

PicoCentauri:
I added the equtions to the docs. Python will render things fine but for C,C++ and Rust currently not. But I think this is something we should treat in a different PR.

python/rascaline/calculators.py
```diff
@@ -155,24 +155,36 @@ def compute(
...
-              **WARNING**: these cell gradients are different from the usual
-              virial used to compute the instantaneous stress tensor (the virial
-              is computed as the gradients of the representations with respect
-              to the strain). To recover the typical virial, you'll need to
-              multiply the cell gradients with the cell matrix.
-
-            The gradients are stored inside the different blocks, and can be
-            accessed with ``descriptor.block(...).gradient(<parameter>)``, where
-            ``<parameter>`` is ``"positions"`` or ``"cell"``.
+              the cell vectors. Cell gradients are compute as
```
ceriottm, line 168: are computed as

python/rascaline/calculators.py
```diff
@@ -155,24 +155,36 @@ def compute(
...
+
+              .. math::
+                  \frac{\partial \langle q \vert A \rangle}
+                       {\partial \mathbf{h}}
+
+              where :math:`\mathbf{h}` is the cell matrix and
+              :math:`\langle q \vert A \rangle` is the representation.
```
ceriottm, line 175: s/is the representation/indicates each of the components of the representation/

python/rascaline/calculators.py
```diff
@@ -155,24 +155,36 @@ def compute(
...
+
+              **Note**: The virial is usually computed as the gradient of the
```
ceriottm, line 177: s/The virial is usually computed as/When computing the virial, one often needs to evaluate/

python/rascaline/calculators.py
```diff
@@ -155,24 +155,36 @@ def compute(
...
+              representation with respect to the strain :math:`\epsilon`.
+              To recover the typical virial from the cell gradient
```
ceriottm, line 179: s/virial/expression/

nb: I'm making all these changes because virial indicates the derivative of the energy, so it may be confusing to call the representation gradient "virial"

rascaline-c-api/include/rascaline.h
```diff
@@ -264,21 +264,33 @@ typedef struct rascal_labels_selection_t {
...
-   *   cell vectors.
+   * - ``"cell"``, for gradients of the representation with respect to
+   *   the cell vectors. Cell gradients are compute as
    *
-   *   **WARNING**: these cell gradients are different from the usual virial
-   *   used to compute the instantaneous stress tensor (the virial is
-   *   computed as the gradients of the representations with respect to the
-   *   strain). To recover the typical virial, you'll need to multiply the
-   *   cell gradients with the cell matrix.
+   *   .. math::
```
ceriottm, line 275: apply also here the changes suggested for `calculators.py`

rascaline-c-api/include/rascaline.hpp
```diff
@@ -454,19 +454,33 @@ class CalculationOptions {
...
-    ///   cell vectors.
+    /// - ``"cell"``, for gradients of the representation with respect to
+    ///   the cell vectors. Cell gradients are compute as
     ///
-    ///   **WARNING**: these cell gradients are different from the usual virial
-    ///   used to compute the instantaneous stress tensor (the virial is
-    ///   computed as the gradients of the representations with respect to the
-    ///   strain). To recover the typical virial, you'll need to multiply the
-    ///   cell gradients with the cell matrix.
+    ///   .. math::
```
ceriottm, line 466: ... and here. Is it really needed to replicate docstrings several times? This seems like a maintenance nightmare.

rascaline/src/calculator.rs
```diff
@@ -154,19 +154,33 @@ impl<'a> LabelsSelection<'a> {
...
-    ///   cell vectors.
+    /// - ``"cell"``, for gradients of the representation with respect to
+    ///   the cell vectors. Cell gradients are compute as
     ///
-    ///   **WARNING**: these cell gradients are different from the usual virial
-    ///   used to compute the instantaneous stress tensor (the virial is
-    ///   computed as the gradients of the representations with respect to the
-    ///   strain). To recover the typical virial, you'll need to multiply the
-    ///   cell gradients with the cell matrix.
+    ///   .. math::
```
ceriottm, line 166: OMG also here.

ceriottm (changes requested):

ceriottm (approved):

PR #110, opened 2022-10-04
Added contribution guidelines

PicoCentauri:
I started adding some information about removing the installed python version of rascalien before writing python docs and ended up refactoring the dev docs a bit:

Basically, I merged the `getting started` and (empty) `contributing` pages. In my opnion both have a big overlap which would end up in linking between them. The contribution guidelines are now also in the well known format `CONTRIBUTING.rst` at the root of the repo.


CONTRIBUTING.rst
```diff
@@ -0,0 +1,195 @@
...
+
+You will need to install and get familiar with the following tools when working
+on rascaline:
+
+- **git**: the software we use for version control of the source code. See
+  https://git-scm.com/downloads for installation instructions.
+- **the rust compiler**: you will need both ``rustc`` (the compiler) and
+  ``cargo`` (associated build tool). You can install both using `rustup`_, or
+  use a version provided by your operating system. We need at least Rust version
+  1.42.0 to build rascaline.
```
Luthaf, line 63: This is a bit outdated, we now require at least version 1.56 (https://github.com/Luthaf/rascaline/blob/de7fb3508935b638166fd51c52f1689438c5cc30/.github/workflows/tests.yml#L34)

CONTRIBUTING.rst
```diff
@@ -0,0 +1,195 @@
...
+    tox -e build  # python packaging
+    tox -e format  # format all files
+
+The latter command `tox -e format` will use tox to do actual formatting instead
+of just testing it.
+
+Writing your own calculator
+---------------------------
+
+For adding a new calculator take a look at :ref:`dev-adding-new-calculator`.
```
Luthaf, line 163: I don't think the link will work from github, should we use an absolute URL instead?

Luthaf (approved):

PR #113, opened 2022-10-10
Fixed some typos

salman-shah2022:


PicoCentauri:
Thanks for fixing these @salman-shah2022. 🥳

We might define if we are using BE or AE for the docs in the future.

Issue #86, opened 2022-07-01
Add examples/tutorials doing sample and feature selection


Luthaf:
All of the code infrastructure is there, but the user-side documentation is missing.

If anyone want to give this a go, let me know!

PR #116, opened 2022-10-13
Added how-to on the topics feature selection and sample selection

hurricane642:
Solution of the #86 

Below are small guides on how to use sample and feature selection in rascaline. The example is made in agreement with a tutorial already available.

### DIFF
--- a/docs/src/how-to/feature-selection.rst
+++ b/docs/src/how-to/feature-selection.rst
@@ -1,4 +1,0 @@
-Feature Selection
-=================
-
-TBD
--- a/python/examples/property-selection.py
+++ b/python/examples/property-selection.py
@@ -1,0 +1,127 @@
+import sys
+
+import chemfiles
+import numpy as np
+from equistore import EquistoreError, Labels, TensorBlock, TensorMap
+
+# we will use the implementation of Farthest Point Sampling from scikit-cosmo,
+# if you want to learn more have a look at the corresponding documentation:
+# <https://scikit-cosmo.readthedocs.io/en/latest/>
+from skcosmo.feature_selection import FPS
+
+from rascaline import SoapPowerSpectrum
+
+
+with chemfiles.Trajectory(sys.argv[1]) as trajectory:
+    frames = [f for f in trajectory]
+
+HYPER_PARAMETERS = {
+    "cutoff": 5.0,
+    "max_radial": 6,
+    "max_angular": 4,
+    "atomic_gaussian_width": 0.3,
+    "center_atom_weight": 1.0,
+    "radial_basis": {
+        "Gto": {},
+    },
+    "cutoff_function": {
+        "ShiftedCosine": {"width": 0.5},
+    },
+}
+
+calculator = SoapPowerSpectrum(**HYPER_PARAMETERS)
+
+descriptor = calculator.compute(frames)
+
+# The selections for feature can be a set of `Labels`, in which case
+# the names of the labels must be a subset of the names of the
+# properties produced by the calculator. You can see the default set
+# of names with:
+print("properties names:", descriptor.property_names)
+
+# We can use a subset of these names to define a selection.
+# In this case, only properties matching the labels in this selection will
+# be used by rascaline (here, only properties with `l=0` will be used)
+selection = Labels(
+    names=["l"],
+    values=np.array([[0]]),
+)
+selected_descriptor = calculator.compute(frames, selected_properties=selection)
+
+selected_descriptor.keys_to_samples("species_center")
+selected_descriptor.keys_to_properties(["species_neighbor_1", "species_neighbor_2"])
+
+properties = selected_descriptor.block().properties
+# we expect to get `[0]` as the list of `l` properties
+print(f"we have the following angular components: {np.unique(properties['l'])}")
+
+# The previous selection method uses the same selection for all blocks.
+# If you can to use different selection for different blocks, you should
+# use a `TensorMap` to create your selection
+
+selected_descriptor = calculator.compute(frames, selected_properties=selection)
+descriptor_for_comparison = calculator.compute(
+    frames, selected_properties=selected_descriptor
+)
+
+# The descriptor had 180 properties stored in the first block,
+# the selected_descriptor had 36. So descriptor_for_comparison
+# will also have 36 properties.
+print("shape of first block initially:", {descriptor.block(0).values.shape})
+print("shape of first block of reference:", {selected_descriptor.block(0).values.shape})
+print(
+    "shape of first block after selection:",
+    {descriptor_for_comparison.block(0).values.shape},
+)
+
+
+# The `TensorMap` format allows us to select different features within
+# each block, and then construct a general matrix of features. We
+# can select the most significant features using FPS, which selects
+# features based on the distance between them. The following code snippet
+# selects the 10 most important features in each block, then constructs
+# a new descriptor based on them, and calculates the final matrix of
+# features for it.
+
+# choose features:
+def fps_feature_selection(descriptor, n_to_select):
+    """
+    Select ``n_to_select`` features block by block in the ``descriptor``, using
+    Farthest Point Sampling to do the selection; and return a ``TensorMap`` with
+    the right structure to be used as properties selection with rascaline calculators
+    """
+    blocks = []
+    for _, block in descriptor:
+        # create a separate FPS selector for each block
+        fps = FPS(n_to_select=n_to_select)
+        mask = fps.fit(block.values).get_support()
+        selected_properties = block.properties[mask]
+        # The only thing that is important in it is the properties,
+        # so we can create samples and components empty altogether.
+        # In the case of values, we only need form matching with len
+        # of properties.
+        # put the selected features in a format rascaline can use.
+        blocks.append(
+            TensorBlock(
+                # values, samples and component carry no information here
+                values=np.empty((1, len(selected_properties))),
+                samples=Labels.single(),
+                components=[],
+                properties=selected_properties,
+            )
+        )
+
+    return TensorMap(descriptor.keys, blocks)
+
+
+# We get `TensorMap` in the results of this function.
+selection = fps_feature_selection(descriptor, n_to_select=10)
+# use the selection with rascaline
+selected_descriptor = calculator.compute(frames, selected_properties=selection)
+
+# Note that in this case it is no longer possible to have a single feature matrix,
+# because each block will have its own properties.
+try:
+    selected_descriptor.keys_to_samples("species_center")
+except EquistoreError:
+    print("Getting Equistore error")
--- a/python/examples/sample-selection.py
+++ b/python/examples/sample-selection.py
@@ -1,0 +1,100 @@
+import sys
+
+import chemfiles
+import numpy as np
+from equistore import Labels
+
+from rascaline import SoapPowerSpectrum
+
+
+with chemfiles.Trajectory(sys.argv[1]) as trajectory:
+    frames = [f for f in trajectory]
+
+HYPER_PARAMETERS = {
+    "cutoff": 5.0,
+    "max_radial": 6,
+    "max_angular": 4,
+    "atomic_gaussian_width": 0.3,
+    "center_atom_weight": 1.0,
+    "radial_basis": {
+        "Gto": {},
+    },
+    "cutoff_function": {
+        "ShiftedCosine": {"width": 0.5},
+    },
+}
+
+calculator = SoapPowerSpectrum(**HYPER_PARAMETERS)
+
+descriptor = calculator.compute(frames)
+
+# The selections for sample can be a set of `Labels`, in which case
+# the names of the labels must be a subset of the names of the
+# samples produced by the calculator. You can see the default set
+# of names with:
+print("samples names:", descriptor.sample_names)
+
+# We can use a subset of these names to define a selection.
+# In this case, only samples matching the labels in this selection
+# will be used by rascaline (here, only atoms from
+# structures 0, 2, and 3)
+selection = Labels(
+    names=["structure"],
+    values=np.array([[0], [2], [3]]),
+)
+
+descriptor_partial_structures = calculator.compute(frames, selected_samples=selection)
+
+descriptor_partial_structures.keys_to_samples("species_center")
+descriptor_partial_structures.keys_to_properties(
+    ["species_neighbor_1", "species_neighbor_2"]
+)
+
+samples = descriptor_partial_structures.block().samples
+# The first block should have `[0, 2, 3]` as `samples["structure"]`
+print(f"we have the following structures: {np.unique(samples['structure'])}")
+
+# If we want to select not only based on the structure indexes but also
+# atomic indexes, we can do the following (here we select atom 0 in the
+# first structure and atom 1 in the third structure):
+selection = Labels(
+    names=["structure", "center"],
+    values=np.array([[0, 0], [2, 1]]),
+)
+
+descriptor_partial_structures = calculator.compute(frames, selected_samples=selection)
+descriptor_partial_structures.keys_to_samples("species_center")
+descriptor_partial_structures.keys_to_properties(
+    ["species_neighbor_1", "species_neighbor_2"]
+)
+
+# The values will have 2 rows, since we have two samples:
+print(
+    "shape of first block of descriptor:",
+    {descriptor_partial_structures.block(0).values.shape},
+)
+
+# The previous selection method uses the same selection for all blocks.
+# If you can to use different selection for different blocks, you should
+# use a `TensorMap` to create your selection
+
+descriptor = calculator.compute(frames)
+descriptor_partial_structures = calculator.compute(frames, selected_samples=selection)
+
+print(type(descriptor_partial_structures))
+# notice how we are passing a TensorMap as the `selected_samples` argument:
+descriptor_for_comparison = calculator.compute(
+    frames, selected_samples=descriptor_partial_structures
+)
+# The descriptor had 420 samples stored in the first block,
+# the descriptor_partial_structures had 0. So descriptor_for_comparison
+# will also have 0 samples.
+print("shape of first block initially:", {descriptor.block(0).values.shape})
+print(
+    "shape of first block of reference:",
+    {descriptor_partial_structures.block(0).values.shape},
+)
+print(
+    "shape of first block after selection:",
+    {descriptor_for_comparison.block(0).values.shape},
+)
--- a/docs/src/how-to/index.rst
+++ b/docs/src/how-to/index.rst
@@ -12,4 +12,4 @@ are a total beginner, you can go to :ref:`userdoc-get-started` section.
 
     computing-soap
     sample-selection
-    feature-selection
+    property-selection
--- a/docs/src/how-to/property-selection.rst
+++ b/docs/src/how-to/property-selection.rst
@@ -1,0 +1,20 @@
+Property Selection
+==================
+
+This examples shows how to only compute a subset of properties for each sample with 
+a given rascaline representation. In particular, we will use the SOAP power spectrum representation, 
+and select the most significant features within a single block using farthest point sampling (FPS). 
+We will run the calculation for all atoms in a structure file, the path to which should be given as the first command line argument.
+
+This is useful if we are interested in the contribution of individual features to the result, 
+or if we want to reduce the computational cost by using only part of the features for our model. 
+
+The first part of this example repeats the :ref:`userdoc-how-to-computing-soap`, so we 
+suggest that you read it initially. 
+
+.. tabs::
+
+    .. group-tab:: Python
+
+        .. literalinclude:: ../../../python/examples/property-selection.py
+            :language: python
--- a/docs/src/how-to/sample-selection.rst
+++ b/docs/src/how-to/sample-selection.rst
@@ -1,4 +1,22 @@
 Sample Selection
 ================
 
-TBD
+This examples shows how to compute a representation only for a subset of the 
+available samples. In particular, we will compute the SOAP power spectrum representation
+for a specific subset of atoms, out of all the atoms in a structure file.
+The path to the structure file is taken from the first command line argument.
+
+This can be useful if we are only interested in certain structures in a large 
+dataset, or if we need to determine the effect of a certain type of atoms on 
+some structure properties. In the following, we will look at the tools with which 
+sample selection can be done in rascaline.
+
+The first part of this example repeats the :ref:`userdoc-how-to-computing-soap`, so we 
+suggest that you read it initially. 
+
+.. tabs::
+
+    .. group-tab:: Python
+
+        .. literalinclude:: ../../../python/examples/sample-selection.py
+            :language: python


python/examples/sample-selection.py
```diff
@@ -0,0 +1,74 @@
...
+
+calculator = SoapPowerSpectrum(**HYPER_PARAMETERS)
+
+descriptor = calculator.compute(frames)
+
+# The names of the samples that will be chosen during training
+# coincide with the samples of the blocks obtained after calculating the
+# properties. Accordingly, the names of the samples can be obtained with the
+# following commands (usually 'structure' (structure number) and 'center'
+# (atom number in the structure).
```
Luthaf, line 34: I don't understand these sentences. What are you trying to say?

python/examples/sample-selection.py
```diff
@@ -0,0 +1,74 @@
...
+print(f"samples names: {descriptor.block(0).samples.names}")
```
Luthaf, line 35:
```suggestion
print("samples names:", descriptor.samples_names)
```

python/examples/sample-selection.py
```diff
@@ -0,0 +1,74 @@
...
+
+# selected_samples can be a variable of `Labels` or `TensorMap` type.
+# In the first case, if we want to select first four structures:
+labels = Labels(
+    names=["structure"],
+    values=np.array([[0], [1], [2], [3]]),
+)
```
Luthaf, line 42: I would not repeat the documentation (`selected_samples can be a variable of `Labels` or `TensorMap` type.`), but explain what is going on and why

```suggestion
# The first way to select samples is with a `Labels` selection. In this case, only samples 
# matching the labels in this selection will be used
selection = Labels(
    names=["structure"],
    values=np.array([[0], [2], [3]]),
)
```

python/examples/sample-selection.py
```diff
@@ -0,0 +1,74 @@
...
+
+reference_descriptor = calculator.compute(frames, selected_samples=labels)
+
+# In the latter case, from the `TensorMap`, which we pass, the list
+# of samples is taken, which will be in the final result. The code
+# with looks as following:
```
Luthaf, line 48:
```suggestion
# The previous selection method uses the same selection for all blocks.
# If you can to use different selection for different blocks, you should use a `TensorMap`
# to create your selection
```

python/examples/sample-selection.py
```diff
@@ -0,0 +1,74 @@
...
+    names=["structure", "center"],
+    values=np.array([[0, 0], [1, 0]]),
+)
+
+descriptor = calculator.compute(frames, selected_samples=labels)
+descriptor.keys_to_samples("species_center")
+descriptor.keys_to_properties(["species_neighbor_1", "species_neighbor_2"])
+
+# the final matrix will have 2 rows, which is what we expected.
+print(f"shape of first block of descriptor: {descriptor.block(0).values.shape}")
```
Luthaf, line 74: I would move this near the other `Labels`-based selection

python/examples/sample-selection.py
```diff
@@ -0,0 +1,74 @@
...
+descriptor_for_comparison = calculator.compute(
+    frames, selected_samples=reference_descriptor
+)
+print(f"shape of first block initially: {descriptor.block(0).values.shape}")
+print(
+    f"shape of first block of reference: {reference_descriptor.block(0).values.shape}"
+)
+print(
+    f"shape of first block after selection: {descriptor_for_comparison.block(0).values.shape}"
+)
```
Luthaf, line 59: This does not really demonstrate the advantage of using a TensorMap for selection (different selections for different keys). Could we do it differently?

python/examples/feature-selection.py
```diff
@@ -0,0 +1,109 @@
...
+# features based on the  distance between them. The following algorithm
+# selects the 10 most important features in each block, then constructs
+# a new descriptor based on them, and calculates the final matrix of
+# features for it.
+
+# choose features:
+features_list = []
+for i in range(len(descriptor)):
+    fps = FPS(n_to_select=10)
+    values = descriptor.block(i).values
```
Luthaf, line 75: maybe the code doing the selection should be moved to a separate function, to make the whole file easier to follow?

python/examples/sample-selection.py
hurricane642: > I don't understand these sentences. What are you trying to say?

That the names of samples to be used for selected_samples are the same as those for blocks in the resulting descriptor

python/examples/feature-selection.py
hurricane642: > maybe the code doing the selection should be moved to a separate function, to make the whole file easier to follow?

I think it is worth explicitly showing exactly how the FPS for each block occurs. If people need to implement it themselves, there's no need to hide parts of the code from them, as I think.

python/examples/sample-selection.py
hurricane642: > This does not really demonstrate the advantage of using a TensorMap for selection (different selections for different keys). Could we do it differently?

Well, we can do feature selection block by block, then do FPS on the whole dataset and see which result is better (for the same number of features in the final matrix). Is that ok?

Luthaf: this example is about sample selection. We could also only have a `TensorMap` selection for features selection, which will be easier to explain

python/examples/feature-selection.py
Luthaf: I don't mean a hidden function, but defining a function inside this example file! This way, people are also able to copy/paste and re-use this function =)

python/examples/sample-selection.py
Luthaf: right, but this is not the case (the names in the selection need to be a subset of the samples names from the calculator, not necessarily the same ones). How about something like

```
The selections for sample can be a set of `Labels`, in which case the names of the labels must be 
a subset of the names of the samples produced by the calculator. You can see the default set of names with:
```

python/examples/feature-selection.py
hurricane642: > I don't mean a hidden function, but defining a function inside this example file! This way, people are also able to copy/paste and re-use this function =)

Ah, okay, I agree!

python/examples/sample-selection.py
hurricane642: Ah, hmm. Well, we can try to output all the data block by block, but that would be more just an illustration of how the algorithm works. About feature_selection - it seems to me that there's nothing wrong with explaining a simple way of working. It doesn't clutter up the tutorial, but it gives some useful background.

python/examples/sample-selection.py
hurricane642: This function doesn't work for me, `samples_names` is not a `TensorMap` attribute. Surely it is already added to the repository? I installed the latest version of equistore, but that doesn't help.

PicoCentauri:
Would these two how to guides not fit better into equistore docs? These looks more like a general selection of Tensormaps to me. Or am I wrong here?

hurricane642:
> Would these two how to guides not fit better into equistore docs? These looks more like a general selection of Tensormaps to me. Or am I wrong here?

It seems to me that the important thing here is that we show selection exactly in the context of the `compute` function of the `rascaline`.

PicoCentauri:
Because you say that the Labels and how we define the element numbers is specific for Rascaline?

On a first glance it looks that the the rascaline specific part, the computation is done. Afterwards, we manipulate the object using equistore methods and nothing rascaline specific.

hurricane642:
> Because you say that the Labels and how we define the element numbers is specific for Rascaline?
> 
> On a first glance it looks that the the rascaline specific part, the computation is done. Afterwards, we manipulate the object using equistore methods and nothing rascaline specific.

The main emphasis is on `selection` (feature or sample) in `rascaline` anyway - I try to give examples of exactly what can be passed as arguments, what the result is, and why these procedures do exactly what is expected. And because `rascaline` is very closely related to `equistore`, I had to do quite a bit of work with it as well.

python/examples/sample-selection.py
```diff
@@ -0,0 +1,113 @@
...
+
+# The selections for sample can be a set of `Labels`, in which case
+# the names of the labels must be a subset of the names of the
+# samples produced by the calculator. You can see the default set
+# of names with:
+print("samples names:", descriptor.sample_names)
+
+# Now we can use these names for selection.
+# In this case, only samples  matching the labels in this selection
+# will be used
```
Luthaf, line 39:
```suggestion
# We can use a subset of these names to define a selection.
# In this case, only samples matching the labels in this selection will be used by rascaline
# (here, only atoms from structures 0, 2, and 3)
```

python/examples/sample-selection.py
```diff
@@ -0,0 +1,113 @@
...
+
+reference_descriptor.keys_to_samples("species_center")
+reference_descriptor.keys_to_properties(["species_neighbor_1", "species_neighbor_2"])
+descriptor.keys_to_samples("species_center")
+descriptor.keys_to_properties(["species_neighbor_1", "species_neighbor_2"])
+
+reference_features = reference_descriptor.block(0).values
+features_direct = descriptor.block(0).values
+features_direct = features_direct[indices]
+print("Are they are equal?", np.array_equal(reference_features, features_direct))
```
Luthaf, line 77: This seems to be a bit convoluted to show that we have something. How about printing the samples for some of the blocks, or grouping the blocks and then doing something like:

```
reference_descriptor.keys_to_samples("species_center")
reference_descriptor.keys_to_properties(["species_neighbor_1", "species_neighbor_2"])

samples = reference_descriptor.block().samples
print(f"we have the following structures: {np.unique(samples['structure'])}")
```

python/examples/sample-selection.py
```diff
@@ -0,0 +1,113 @@
...
+
+# If we want to select not only numbers of structures, but also
+# numbers of atoms, we can do the following (to select the first atoms
+# in the first two structures)
```
Luthaf, line 81:
```suggestion
# If we want to select not only based on the structure indexes but also
# atomic indexes, we can do the following: (here we select the first atom
# in the first two structures)
```

python/examples/sample-selection.py
```diff
@@ -0,0 +1,113 @@
...
+selection = Labels(
+    names=["structure", "center"],
+    values=np.array([[0, 0], [1, 0]]),
+)
+
+reference_descriptor = calculator.compute(frames, selected_samples=selection)
+reference_descriptor.keys_to_samples("species_center")
+reference_descriptor.keys_to_properties(["species_neighbor_1", "species_neighbor_2"])
+
+# the final matrix will have 2 rows, which is what we expected.
```
Luthaf, line 91:
```suggestion
# the values will have 2 rows, which is what we expected.
```

python/examples/sample-selection.py
```diff
@@ -0,0 +1,113 @@
...
+    frames, selected_samples=reference_descriptor
+)
+print("shape of first block initially:", {descriptor.block(0).values.shape})
+print(
+    "shape of first block of reference:", {reference_descriptor.block(0).values.shape}
+)
+print(
+    "shape of first block after selection:",
+    {descriptor_for_comparison.block(0).values.shape},
+)
```
Luthaf, line 113: This does demonstrate the ability to use TensorMap, but feel a bit forced. I don't have a good idea on how to replace it, so maybe we should leave this out of this example? Something similar is being done for the properties selection.

docs/src/how-to/feature-selection.rst
```diff
@@ -1,4 +1,21 @@
 Feature Selection
```
Luthaf, line 1: This should be renamed (this file, the example file and titles) to refer to `properties`, since that's the name equistore uses.

docs/src/how-to/feature-selection.rst
```diff
@@ -1,4 +1,21 @@
...
 =================
 
-TBD
\ No newline at end of file
+This examples shows how to compute certain SOAP power spectrum descriptor 
+properties for each atoms of each structures of a provided structure file, as well as 
+a methodology for selecting the most significant features with usage of farthest point sampling (FPS)
+algorithm within a single block. The path to the structure file is taken from the first command line
+argument.
```
Luthaf, line 8:
```suggestion
This examples shows how to only compute a subset of properties for each sample with 
a given rascaline representation. In particular, we will use the SOAP power spectrum representation, 
and select the most significant features within a single block using farthest point sampling (FPS). 
We will run the calculation for all atoms in a structure file, the path to which should be given as the first command line argument.
```

docs/src/how-to/feature-selection.rst
```diff
@@ -1,4 +1,21 @@
...
+
+This can be useful if we are interested in the contribution of individual features to the result, 
+or if we want to reduce the computational cost by using only part of the features for our model.
```
Luthaf, line 11:
```suggestion
This is useful if we are interested in the contribution of individual features to the result, 
or if we want to reduce the computational cost by using only part of the features for our model. 
```

python/examples/feature-selection.py
```diff
@@ -0,0 +1,116 @@
+import sys
+
+import chemfiles
+import numpy as np
+from equistore import Labels, TensorBlock, TensorMap
+from skcosmo.feature_selection import FPS
```
Luthaf, line 6: This is worth a comment, maybe 

```
# we will use the implementation of Farthest Point Sampling from scikit-cosmo, 
# if you want to learn more have a look at the corresponding documentation: <link>
from skcosmo.feature_selection import FPS
```

python/examples/feature-selection.py
```diff
@@ -0,0 +1,116 @@
...
+
+# The selections for feature can be a set of `Labels`, in which case
+# the names of the labels must be a subset of the names of the
+# properties produced by the calculator. You can see the default set
+# of names with:
+print("properties names:", descriptor.property_names)
+
+# Now we can use these names for selection.
+# In this case, only features  matching the labels in this selection
+# will be used
```
Luthaf, line 40:
```suggestion
# We can use a subset of these names to define a selection.
# In this case, only properties matching the labels in this selection will be used by rascaline
# (here, only `l=0` components will be used)
```

python/examples/feature-selection.py
```diff
@@ -0,0 +1,116 @@
...
+selection = Labels(
+    names=["l"],
+    values=np.array([[0]]),
+)
+reference_descriptor = calculator.compute(frames, selected_properties=selection)
```
Luthaf, line 45:
```suggestion
selected_descriptor = calculator.compute(frames, selected_properties=selection)
```

python/examples/feature-selection.py
```diff
@@ -0,0 +1,116 @@
...
+
+# The previous selection method uses the same selection for all blocks.
+# If you can to use different selection for different blocks, you should
+# use a `TensorMap` to create your selection
```
Luthaf, line 49: This comment should move after the print code below since it refers to stuff happening later

python/examples/feature-selection.py
```diff
@@ -0,0 +1,116 @@
...
+            ]
+        )
+    property_values = np.asarray(property_values).reshape(-1, 3)
+    properties = Labels(descriptor.block(i).properties.names, property_values)
+    set_of_blocks.append(
+        TensorBlock(
+            values=values, samples=samples, components=components, properties=properties
+        )
+    )
+new_descriptor = TensorMap(descriptor.keys, set_of_blocks)
```
Luthaf, line 111: This is manually re-doing what rascaline can do for you. I would do something like
```suggestion
# choose features:
def fps_feature_selection(descriptor, n_to_select):
    blocks = []
    for _, block in descriptor:
        # create a separate FPS selector for each block
        fps = FPS(n_to_select=n_to_select)
        mask = fps.fit(block.values).get_support()
        selected_properties = block.properties[mask]

        # put the selected features in a format rascaline can use
        blocks.append(TensorBlock(
            # values, samples and component carry no information here
            values=np.empty(),
            samples=Labels.single(),
            components=[],
            properties=selected_properties
        ))

        
    return TensorMap(descriptor.keys, blocks)


selection = fps_feature_selection(descriptor, n_to_select=10)

# use the selection with rascaline
selected_descriptor = calculator.compute(frames, selected_properties=selection)
```

python/examples/sample-selection.py
```diff
@@ -0,0 +1,113 @@
...
+
+# Now we can use these names for selection.
+# In this case, only samples  matching the labels in this selection
+# will be used
+selection = Labels(
+    names=["structure"],
+    values=np.array([[0], [2], [3]]),
+)
+
+reference_descriptor = calculator.compute(frames, selected_samples=selection)
```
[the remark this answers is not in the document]
hurricane642: Ok, sure, I was also thinking about what to call it better.

python/examples/sample-selection.py
hurricane642: True yes, it's much easier that way

python/examples/sample-selection.py
hurricane642: On the one hand yes, on the other hand, I personally find it very annoying when some tutorial doesn't show all the functionality. It's helpful at least to show how the function works in the case of `TensorMap`

PicoCentauri (approved):
I am happy with the two how-tos and would keep them for now. I will create an upcoming PR where I port these examples to produce actual outputs as we are doing this for the introduction tutorials (see #119)

We can then further discuss details if we like.

python/examples/sample-selection.py
```diff
@@ -0,0 +1,99 @@
...
+descriptor_partial_structures = calculator.compute(frames, selected_samples=selection)
+
+descriptor_partial_structures.keys_to_samples("species_center")
+descriptor_partial_structures.keys_to_properties(
+    ["species_neighbor_1", "species_neighbor_2"]
+)
+
+samples = descriptor_partial_structures.block().samples
+# The first block will have and array([0,2,3]) as a `samples`
+# which is to be expected.
```
Luthaf, line 55:
```suggestion
# The first block should have `[0, 2, 3]` as `samples["structure"]`
```

python/examples/sample-selection.py
```diff
@@ -0,0 +1,99 @@
...
+print(f"we have the following structures: {np.unique(samples['structure'])}")
+
+# If we want to select not only based on the structure indexes but also
+# atomic indexes, we can do the following: (here we select the first atom
+# in the first two structures).
```
Luthaf, line 60:
```suggestion
# atomic indexes, we can do the following (here we select atom 0 in the 
# first structure and atom 1 in the third structure):
```

python/examples/sample-selection.py
```diff
@@ -0,0 +1,99 @@
...
+    values=np.array([[0, 0], [2, 1]]),
+)
+
+descriptor_partial_structures = calculator.compute(frames, selected_samples=selection)
+descriptor_partial_structures.keys_to_samples("species_center")
+descriptor_partial_structures.keys_to_properties(
+    ["species_neighbor_1", "species_neighbor_2"]
+)
+
+# The values will have 2 rows, which is what we expected.
```
Luthaf, line 72:
```suggestion
# The values will have 2 rows, since we have two samples:
```

python/examples/sample-selection.py
```diff
@@ -0,0 +1,99 @@
...
+    {descriptor_partial_structures.block(0).values.shape},
+)
+
+# The previous selection method uses the same selection for all blocks.
+# If you can to use different selection for different blocks, you should
+# use a `TensorMap` to create your selection
+
+descriptor = calculator.compute(frames)
+descriptor_partial_structures = calculator.compute(frames, selected_samples=selection)
+
```
Luthaf, line 84:
```suggestion

print(type(descriptor_partial_structures))
# notice how we are passing a TensorMap as the `selected_samples` argument:
```

python/examples/property-selection.py
```diff
@@ -0,0 +1,121 @@
...
+
+# The selections for feature can be a set of `Labels`, in which case
+# the names of the labels must be a subset of the names of the
+# properties produced by the calculator. You can see the default set
+# of names with:
+print("properties names:", descriptor.property_names)
+
+# We can use a subset of these names to define a selection.
+# In this case, only properties matching the labels in this selection will
+# be used by rascaline (here, only `l=0` components will be used)
```
Luthaf, line 44: (components refer to something different)

```suggestion
# be used by rascaline (here, only properties with `l=0` will be used)
```

python/examples/property-selection.py
```diff
@@ -0,0 +1,121 @@
...
+    names=["l"],
+    values=np.array([[0]]),
+)
+selected_descriptor = calculator.compute(frames, selected_properties=selection)
+
+selected_descriptor.keys_to_samples("species_center")
+selected_descriptor.keys_to_properties(["species_neighbor_1", "species_neighbor_2"])
+
+properties = selected_descriptor.block().properties
+# get an array([0]) as a list of angular properties
```
Luthaf, line 55:
```suggestion
# we expect to get `[0]` as the list of `l` properties
```

python/examples/property-selection.py
```diff
@@ -0,0 +1,121 @@
...
+# will also have 36 properties.
+print("shape of first block initially:", {descriptor.block(0).values.shape})
+print("shape of first block of reference:", {selected_descriptor.block(0).values.shape})
+print(
+    "shape of first block after selection:",
+    {descriptor_for_comparison.block(0).values.shape},
+)
+
+
+# The `TensorMap` format allows us to select features within
```
Luthaf, line 78:
```suggestion
# The `TensorMap` format allows us to select different features within
```

python/examples/property-selection.py
```diff
@@ -0,0 +1,121 @@
...
+# each block, and then construct a general matrix of features. We
+# can select the most significant features using FPS, which selects
+# features based on the  distance between them. The following algorithm
```
Luthaf, line 81:
```suggestion
# features based on the distance between them. The following code snippet
```

python/examples/property-selection.py
```diff
@@ -0,0 +1,121 @@
...
+# selects the 10 most important features in each block, then constructs
+# a new descriptor based on them, and calculates the final matrix of
+# features for it.
+
+# choose features:
+def fps_feature_selection(descriptor, n_to_select):
+    blocks = []
```
Luthaf, line 88:
```suggestion
def fps_feature_selection(descriptor, n_to_select):
    """ 
    Select ``n_to_select`` features block by block in the ``descriptor``, using 
    Farthest Point Sampling to do the selection; and return a ``TensorMap`` with
    the right structure to be used as properties selection with rascaline calculators
    """
    blocks = []
```

python/examples/property-selection.py
```diff
@@ -0,0 +1,121 @@
...
+            )
+        )
+
+    return TensorMap(descriptor.keys, blocks)
+
+
+# We get `TensorMap` in the results of this function. The only thing that
+# is important in it is the properties, so we can create samples and
+# components empty altogether. In the case of values, we only need
+# form matching with len of properties.
```
Luthaf, line 111: I would move this comment (only the properties are important, we leave the rest empty) inside the function above

python/examples/property-selection.py
```diff
@@ -0,0 +1,121 @@
...
+selection = fps_feature_selection(descriptor, n_to_select=10)
+# use the selection with rascaline
+selected_descriptor = calculator.compute(frames, selected_properties=selection)
+
+# Note that in this case it is no longer possible to have a single feature matrix,
+# because each block will have its own properties.
```
Luthaf, line 117:
```suggestion
# because each block will have its own set of properties.
```

also, we should actually be able to move keys around, but it would require changing the names/values of the properties from (n1, n2, l) to (q) and [0, 1, 2, 3, ..., 9] instead of the actual values. Not sure if it is worth doing in this how to?

hurricane642: This is an interesting question. On the one hand, people may have such a problem and it would be good to show them the solution. On the other hand, the tutorial already looks a bit large, and it might be worth writing another tutorial looking at the extra propblems there.

Luthaf: Sure, let's leave it for later!

Luthaf:
Thanks a lot @hurricane642!

PR #117, opened 2022-10-14
Extract the spline code into a standalone module

Luthaf:
This make the spline re-usable across the project, and changes their output to an arbitrary array shape, opening the door to different radial resolution per angular channel in SOAP.


PicoCentauri (approved):
LGTM

PR #114, opened 2022-10-11
Added digamma and hyp2f1 functions

PicoCentauri:
The code has to be cleaned up and some of the tests have to be fixed or values excluded.

PR #118, opened 2022-10-18
Update CI

Luthaf:
Github Actions is deprecating ubuntu-18.04


PicoCentauri:
@Luthaf you were quite quick with merging this. It seems that there are unwanted files py files from #114  in this PR.

We could also use `ubuntu-latest` instead of a hardcoded version...

Luthaf:
cr*p ... I'll clean up master immediatly

PR #119, opened 2022-10-18
Added tutorials for new users

PicoCentauri:
* Added two tutorials for new users using sphinx galleries
* Update equations on cutoff functions to be consistent with the tutorial

One question might be where to put the tutorials? I would prefer to put them next to the `python/examples` at `python/tutorials` but this might be hard with loading the explicit dataset...

# TODO
- [x] Check if the hypers are reasonable to be copy pasted.


docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,148 @@
+"""
```
Luthaf, line 1: could you add these files to the Python lint path?

PicoCentauri: sure, but we might thought about moving them to `python/tutorials`, which would be consistent with the examples folder. However, then I am not sure how to cleverly handle the dataset withot doing some extensive linking...

Luthaf:
The doc build failure is because we don't install rascaline when creating the docs (since all the doc can be extracted from Python without loading `librascaline.so`). You'll need to install rascaline around here: https://github.com/Luthaf/rascaline/blob/275e5d2916126e7aa10a38ecbcea7c3fe663ecb3/.github/workflows/docs.yml#L26-L29, and remove the extra PYTHONPATH entry here: https://github.com/Luthaf/rascaline/blob/275e5d2916126e7aa10a38ecbcea7c3fe663ecb3/docs/src/conf.py#L12

PicoCentauri:
Using `reference_url` in the `sphinx_gallery_conf` should create hyperlinks in the example scripts linking the used functions to their matching online documentation. However these [currently does not work](https://github.com/sphinx-gallery/sphinx-gallery/issues/967). I will investigate this further...

PicoCentauri:
I finished the tutorials and maybe @hurricane642 and @Luthaf you can take a look. Still open but we can fix this later are the hyperlinks in code, but we can also fix this later.

docs/src/get-started/tutorials/understanding_hypers.py
```diff
@@ -0,0 +1,405 @@
...
+    print(
+        f"Descriptor for center_weight={center_weight}: {descriptor.block(0).values[0]}"
+    )
+
+# %%
+# Choosing the ``cutoff_function``
+# --------------------------------
+#
+# In a naive descriptor approach all atoms within the cutoff are taken in into
+# account equually and atoms withput the cutoff are ignored. This behaviour
```
hurricane642, line 157:
```suggestion
# account equally and atoms without the cutoff are ignored. This behavior
```

hurricane642 (commented):
It looks great. I've corrected typos and added a couple of links that might be useful. I think it's perfect for a first introduction.

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,161 @@
...
+# %%
+# The descriptor format is a :class:`equistore.tensor.TensorMap` object.
+# Equistore can be like Numpy for storing representation of atomistic ML data.
+# Extensive details on the equistore are be covered in the docs of
+# `equistore <https://lab-cosmo.github.io/equistore/>`_.
+# We now access the actual descriptor data atrributed to the
+# :class:`equistore.tensor.TensorMap` object.
+#
+# The :class:`equistore.tensor.TensorMap` is structured in so called
+# ``blocks``. One block for each angular channel
```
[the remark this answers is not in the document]
PicoCentauri: Really good point!

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,161 @@
...
+    f"{counts[1]} {elements[1]}-atoms, "
+    f"{counts[2]} {elements[2]}-atoms and "
+    f"{counts[3]} {elements[3]}-atoms."
+)
+
+# %%
+# Calculate a descriptor
+# ----------------------
+#
+# We will now calculate the atomic descriptor for this structure using the SOAP spherical expansion.
```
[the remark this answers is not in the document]
PicoCentauri: Thanks, yes I added this!

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,161 @@
...
+# For doing so we have to define a set parameters telling the rascaline how "detailed" the
+# spherical expansion should be calculated. These parameters are also called hyper parameters because
+# they are not being optimized in a usual machine learning procedure.
+# Hyper parameters are the crucial part of calculating descriptors. Poorly selected hyper parameters will
```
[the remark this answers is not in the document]
PicoCentauri: Added

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,161 @@
...
+from rascaline import SphericalExpansion
+
+
+with chemfiles.Trajectory("../../../static/datasets.xyz") as trajectory:
+    frames = [f for f in trajectory]
+
+print(f"The dataset contains {len(frames)} frames.")
+
+# %%
+# Here, we will not explain how to use chemfiles
```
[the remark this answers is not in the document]
PicoCentauri: Good point. I added your paragraph.

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,172 @@
...
+# %%
+# Calculate a descriptor
+# ----------------------
+#
+# We will now calculate the atomic descriptor for this structure using the
+# SOAP spherical expansion as introduced by
+# `Bartók, Kondor and Csányi <http://dx.doi.org/10.1103/PhysRevB.87.184115>`_
+# For doing so we have to define a set parameters telling the rascaline how "detailed" the
+# spherical expansion should be calculated. These parameters are also called hyper parameters because
+# they are not being optimized in a usual machine learning procedure.
```
Luthaf, line 74: not sure about this: one can use bayesian optimization or other technics to optimize the hypers. In general I tend to see hyper-parameters as the parameters of the representation, in opposition to parameters of the models. This breaks when dealing with NN, which mix the two.

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,172 @@
...
+# Hyper parameters are the crucial part of calculating descriptors. Poorly selected hyper parameters will
+# lead to a poor description of your dataset as discussed in
+# `literature <https://arxiv.org/abs/1502.02127>`_.
+#
+# Below are our default hyper parameters for this dataset.
+# The effect and of changing some hyper parameters are discussed in a different tutorial.
```
Luthaf, line 80:
```suggestion
# The effect of changing some hyper parameters are discussed in a different tutorial.
```

And add a link to the other tutorial?

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,172 @@
...
+
+HYPER_PARAMETERS = dict(
```
Luthaf, line 82: I personally prefer using literal for dictionaries instead of the `dict` function, especially here where we have nested dictionaries

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,172 @@
...
+# After we set the hyper parameters we initilize a calculator object linking the the hyper parameters
+# with the a descriptor method.
+
+calculator = SphericalExpansion(**HYPER_PARAMETERS)
+descriptor0 = calculator.compute(frame0)
+print(type(descriptor0))
+
+# %%
+# The descriptor format is a :class:`equistore.tensor.TensorMap` object.
+# Equistore can be like Numpy for storing representation of atomistic ML data.
```
Luthaf, line 103:
```suggestion
# Equistore is like Numpy for storing representation of atomistic ML data.
```

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,172 @@
...
+# Extensive details on the equistore are be covered in the docs of
+# `equistore <https://lab-cosmo.github.io/equistore/>`_.
```
Luthaf, line 105:
```suggestion
# Extensive details on the equistore are be covered in the 
# `corresponding documentation <https://lab-cosmo.github.io/equistore/>`_.
```

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,172 @@
...
+# We now access the actual descriptor data atrributed to the
+# :class:`equistore.tensor.TensorMap` object.
```
Luthaf, line 107:
```suggestion
#
# We will now have a look at how the data is stored inside
# :class:`equistore.tensor.TensorMap` objects.
```

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+print(descriptor0)
+
+# %%
+#
+# The :class:`equistore.tensor.TensorMap` is structured in so called
+# ``blocks``. One block for each angular channel
+# and a combination of two elements labeled by their atomic number, e.g. 1 for
+# hydrogen, 6 for carbon etc.
```
Luthaf, line 117:
```suggestion
# The :class:`equistore.tensor.TensorMap` is structured in
# ``blocks``, each associated with a key. Here we have one block for each angular channel, 
# central atom species and neighbor atom species.
# Different atomic species are represented using their atomic number, e.g. 1 for
# hydrogen, 6 for carbon etc.
```

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,172 @@
...
+# Our descriptor contains 112 blocks covering
+# all combinations of the angular channels and the the elements.
+#
+# Lets take a look at the first block in details. The 0th block contains the
+# descriptor for the 0th angular channel between the hydrogen-hydrogen pairs.
+# The values are stored in an array where the first dimension is 8, because we have 8
+# hydrogen atoms in our frame. The last dimensions represents the number
+# of radial channels. We choose 9 above so there are 9 values in the last dimensions.
+
+print(descriptor0.block(0).values.shape)
```
Luthaf, line 127:
```suggestion
print(descriptor0.keys[0])
print(descriptor0.block(0).values.shape)
```

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,172 @@
...
+# Rascaline is also able to process more than one structure within one function
+# call. You can process a whole dataset with
+
+descriptor_full = calculator.compute(frames)
+print(descriptor_full.block(0).values.shape)
+
+# %%
+# Now, the 0th block of the :class:`equistore.tensor.TensorMap` contains not 8
+# but 420 entries in the first dimensions.
+# This reflects the fact that in total we have 420 H–H pairs in the whole dataset.
```
Luthaf, line 148:
```suggestion
# This reflects the fact that in total we have 420 H atoms in the whole dataset.
```

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,172 @@
...
+# check out the :ref:`userdoc-references` section. Details how to
+# calculate the features using another
+# programming language like Rust or C/C++ are given in the :ref:`userdoc-how-to`.
+#
+# Computing gradients
+# -------------------
+#
+# Additionally, rascaline is also able to calculate the gradients with respect to positions, which is
+# useful for constructing a ML potential. `gradients` can be calculated by setting the
+# `gradients=["positions"]` parameter of the compute method.
```
Luthaf, line 161:
```suggestion
# useful for constructing a ML potential. ``gradients`` can be calculated by setting the
# ``gradients=["positions"]`` parameter of the compute method.
```

This is still using MD syntax

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,172 @@
...
+#
+# If you want to use another calculator as the ``SphericalExpansion`` used here
+# check out the :ref:`userdoc-references` section. Details how to
+# calculate the features using another
+# programming language like Rust or C/C++ are given in the :ref:`userdoc-how-to`.
+#
+# Computing gradients
+# -------------------
+#
+# Additionally, rascaline is also able to calculate the gradients with respect to positions, which is
```
Luthaf, line 159: should we mention gradients w.r.t. cell here as well?

docs/src/get-started/tutorials/get-started.py
hurricane642: It seems to me that it would be best to define these parameters as "model parameters to be configured before the learning process". The most general and honest definition.

docs/src/get-started/tutorials/get-started.py
PicoCentauri: I personally prefer the other, because it easily allows copying between parameters. But, I am fine with changing this.

docs/src/get-started/tutorials/get-started.py
PicoCentauri: I thought about is as well. I maybe makes sense to mention it. I will write a sentence.

docs/src/get-started/tutorials/get-started.py
PicoCentauri: I would be fine with calling them model parameters but the slang word unfortunately is hyper parameters.

Also we can remove the sentence about the optimization. This does not really help a new user here.

docs/src/conf.py
```diff
@@ -105,6 +105,14 @@ def setup(app):
 autodoc_member_order = "bysource"
 autodoc_typehints_format = "short"
 
+sphinx_gallery_conf = {
+    "filename_pattern": "/*",
+    "examples_dirs": ["get-started/tutorials/"],
+    "gallery_dirs": ["get-started/tutorials_rendered"],
+    "min_reported_time": 60,
+    "reference_url": {"rascaline": None},
```
Luthaf, line 113: So, it is possible to have automatic linking of the rascaline reference as well?

docs/src/get-started/tutorials/understanding_hypers.py
```diff
@@ -0,0 +1,405 @@
...
+# compared to atoms that are
+# closer. This can be achieved by a ``radial_scaling`` function
+# with a long-range algebraic decay and smooth behavior at :math:`r \rightarrow 0`
+#
+# .. math::
+#             u(r) = \begin{cases}
+#             \frac{1}{(r/r_0)^m} &\text{if c=0,}\\
+#             1 &\text{if m=0,} \\
+#             \frac{c}{c+(r/r_0)^m} &\text{else},
+#             \end{cases}
```
Luthaf, line 241:
```suggestion
#             u(r) = \begin{dcases}
#             \frac{1}{(r/r_0)^m} &\text{if c=0,}\\
#             1 &\text{if m=0,} \\
#             \frac{c}{c+(r/r_0)^m} &\text{else},
#             \end{dcases}
```

if possible?

docs/src/get-started/tutorials/understanding_hypers.py
```diff
@@ -0,0 +1,405 @@
...
+        return 1 / (r / scale) ** exponent
+    if exponent == 0:
+        return 1
+    else:
+        return rate / (rate + (r / scale) ** exponent)
+
+
+# %%
+# In the following we show three different radial scaling functions, where the
+# first one are our default values.
```
Luthaf, line 278: I'm not fond of "default values" here. There are no default values for a dataset, it depends on the task at hand.

docs/src/get-started/tutorials/understanding_hypers.py
```diff
@@ -0,0 +1,405 @@
...
+
+descriptor_rs = calculator_rs.compute(frame0)
+
+print(f"cosine switching + radial scaling: {str(descriptor_rs.block(0).values[0]):>50}")
+
+# %%
+# Note that the ``radial_scaling`` using a a long-range algebraic decay
+# in rascaline can be chosen by the ``'Willatt2018'`` since this kind of
+# was first introduced in the corresponding
+# `publication <https://doi.org/10.1039/C8CP05921G>`_.
```
Luthaf, line 405:
```suggestion
# in rascaline can be chosen by the ``'Willatt2018'``, corresponding to the 
# function introduced in the corresponding
# `publication <https://doi.org/10.1039/C8CP05921G>`_.
```

docs/src/get-started/tutorials.rst
```diff
@@ -0,0 +1,3 @@
+.. _userdoc-tutorials:
+
+.. include:: tutorials_rendered/index.rst
```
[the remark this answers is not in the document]
PicoCentauri: Yes, I already did this in my last commit.

docs/src/conf.py
PicoCentauri: It should but currently does not work. I will take a another look today.

docs/src/get-started/tutorials/understanding_hypers.py
```diff
@@ -0,0 +1,405 @@
...
+
+# %%
+# For a ``cutoff`` of 0.1 Å there is no neighboring atom within the cutoff and
+# one could expect the all features are 0. However,
+# even here the features finite. This originates from the fact that there is
+# center atom contribution to the descriptor. We can vary this contribution
+# using the ``center_atom_weight`` parameter so that the descriptor
+# finally is 0 every where.
+#
+# Note that the ``center_atom_weight`` can have an influence for some ML model.
```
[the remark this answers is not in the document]
PicoCentauri: TBH, For SOAP I don't know how it influences the model. But, at some point we should add something here. I will leave a comment.

docs/src/get-started/tutorials/understanding_hypers.py
PicoCentauri: Yes, I agree. Even though we should give new users something in hand that they remember. Nothing is more confusing than always saying here a bunch of parameters but do not use them, because they might be wrong.

docs/src/get-started/tutorials/get-started.py
```diff
@@ -0,0 +1,180 @@
...
+#
+# To do so we have to define a set of parameters telling the rascaline how the
+# spherical expansion should be calculated. These parameters are also called hyper parameters because
+# they are not being optimized in a usual machine learning procedure.
+# Hyper parameters are the crucial part of calculating descriptors. Poorly selected hyper parameters will
+# lead to a poor description of your dataset as discussed in
+# `literature <https://arxiv.org/abs/1502.02127>`_.
+#
+# Below are our default hyper parameters for this dataset.
+# The effect of changing some hyper parameters are discussed in a different tutorial.
```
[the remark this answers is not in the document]
PicoCentauri: Done.

docs/src/conf.py
PicoCentauri: I looked in other libraries and our config looks fine. Maybe we take a look together @Luthaf.

PicoCentauri:
Thanks @Luthaf. Everything looks good to me now.

docs/src/get-started/tutorials/understanding-hypers.py
```diff
@@ -171,12 +172,12 @@
 # .. math::
 #
 #     f(r) = \begin{cases}
-#           1 &r < r_c - s,\\
-#           0.5 + 0.5 \cos[\pi (r - r_c + s) / s] &r_c - s < r <= r_c, \\
+#           1 &r < r_c - w,\\
```
PicoCentauri, line 175: If we change it here then also in he doc string of the rust function.

Luthaf: So I undid the changes in the rust docstring, since these are also used here: https://luthaf.fr/rascaline/latest/references/calculators/spherical-expansion.html, and were not rendering properly.

docs/src/get-started/tutorials/understanding-hypers.py
```diff
@@ -45,8 +45,9 @@
 # parameters into the ``SphericalExpansion`` class without defining a
 # ``HYPERPARAMETERS`` dictionary like we did in the previous tutorial.
 
+cutoff = 4.5
 calculator_ext = SphericalExpansion(
-    cutoff=6.0,
+    cutoff=cutoff,
```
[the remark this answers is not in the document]
Luthaf: sure, I'll change it back

PicoCentauri:
I did a final proof read of the two tutorials. If somebody else want to give it a final read please go for it. 

If no, we can ship 🚢 this.

docs/src/get-started/tutorials/first-calculation.py
```diff
@@ -0,0 +1,210 @@
...
+# %%
+#
+# Calculate a descriptor
+# ----------------------
+#
+# We will now calculate an atomic descriptor for this structure using the SOAP
+# spherical expansion as introduced by `Bartók, Kondor, and Csányi
+# <http://dx.doi.org/10.1103/PhysRevB.87.184115>`_.
+#
+# To do so we below define a set of parameters telling rascaline how the
```
Luthaf, line 80:
```suggestion
# To do so we define below a set of parameters telling rascaline how the
```

docs/src/get-started/tutorials/first-calculation.py
```diff
@@ -0,0 +1,210 @@
...
+# object with hyper parameters defined above.
+
+calculator = SphericalExpansion(**HYPER_PARAMETERS)
+descriptor0 = calculator.compute(frame0)
+print(type(descriptor0))
+
+# %%
+#
+# The descriptor format is a :class:`equistore.tensor.TensorMap` object.
+# Equistore is like Numpy for storing a representation of atomistic ML data.
```
Luthaf, line 114:
```suggestion
# Equistore is like numpy for storing representations of atomistic ML data.
```

docs/src/get-started/tutorials/first-calculation.py
```diff
@@ -0,0 +1,210 @@
...
+# %%
+#
+# Now, the 0th block of the :class:`equistore.tensor.TensorMap` contains not 8
+# but 420 entries in the first dimensions. This reflects the fact that in total
+# we have 420 hydrogen atoms in the whole dataset.
+#
+#
+# If you want to use another calculator as the
+# :class:`rascaline.calculators.SphericalExpansion` shown here
+# check out the :ref:`userdoc-references` section.
```
Luthaf, line 185:
```suggestion
# If you want to use another calculator instead of
# :class:`rascaline.calculators.SphericalExpansion` shown here
# check out the :ref:`userdoc-references` section.
```

docs/src/get-started/tutorials/understanding-hypers.py
```diff
@@ -0,0 +1,417 @@
...
+        atomic_gaussian_width=0.3,
+        center_atom_weight=1.0,
+        radial_basis={"Gto": {"spline_accuracy": 1e-6}},
+        cutoff_function={"ShiftedCosine": {"width": 0.5}},
+        radial_scaling={"Willatt2018": {"scale": 2.0, "rate": 1.0, "exponent": 4}},
+    )
+
+descriptor = calculator_cutoff.compute(frame0)
+
+print(f"Descriptor for cutoff={cutoff} Å: {descriptor.block(0).values[0]}")
```
Luthaf, line 128: Shouldn't this stay inside the loop?

docs/src/get-started/tutorials/understanding-hypers.py
```diff
@@ -0,0 +1,417 @@
...
+
+# %%
+#
+# For a ``cutoff`` of 0.1 Å there is no neighboring atom within the cutoff and
+# one could expect all features are 0. This is not the case because the
```
Luthaf, line 133:
```suggestion
# one could expect all features to be 0. This is not the case because the
```

docs/src/get-started/tutorials/understanding-hypers.py
```diff
@@ -0,0 +1,417 @@
...
+        radial_basis={"Gto": {"spline_accuracy": 1e-6}},
+        cutoff_function={"ShiftedCosine": {"width": 0.5}},
+        radial_scaling={"Willatt2018": {"scale": 2.0, "rate": 1.0, "exponent": 4}},
+    )
+
+descriptor = calculator_cutoff.compute(frame0)
+
+print(
+    f"Descriptor for center_weight={center_weight}: " f"{descriptor.block(0).values[0]}"
+)
```
Luthaf, line 157: Shouldn't this stay inside the loop?

docs/src/get-started/tutorials/understanding-hypers.py
```diff
@@ -0,0 +1,417 @@
...
+# %%
+#
+# Choosing the ``radial_scaling``
+# -------------------------------
+#
+# As mentioned above all atoms within the cutoff are taken equally for a
+# descriptor. This might limit an ML model's accuracy for atomistic systems on top
+# of a ``cutoff_function``, so it is sometimes useful to weigh neighbors that
+# further away from the central atom less than neighbors closer to the central
+# atom. This can be achieved by a ``radial_scaling`` function with a long-range
```
Luthaf, line 246: The sentence is a bit unclear, how about

```suggestion
# descriptor. This might limit the accuracy of a model, so it is sometimes useful to weigh neighbors that
# further away from the central atom less than neighbors closer to the central
# atom. This can be achieved by a ``radial_scaling`` function with a long-range
```

docs/src/get-started/tutorials/understanding-hypers.py
```diff
@@ -0,0 +1,417 @@
...
+    if exponent == 0:
+        return 1
+    else:
+        return rate / (rate + (r / scale) ** exponent)
+
+
+# %%
+#
+# In the following we show three different radial scaling functions, where the
+# first one is the parameter we use for the feature calculation in the
```
Luthaf, line 295:
```suggestion
# first one uses the parameters we use for the calculation of features in the
```

docs/src/get-started/tutorials/understanding-hypers.py
PicoCentauri: Ohh yes I will fix this.

docs/src/get-started/tutorials/understanding-hypers.py
PicoCentauri: Thanks!

PicoCentauri:
Thanks for your comments. I implemented them and rebased the history.

Luthaf:
I've added the tutorials to tox, so they are now checked. I think this is now good to go!

PicoCentauri:
Okay, though the files will be run, when we create the docs.

I am fine with this and if the CI is through we can merge.

Luthaf:
Good point, I forgot about that. I'll remove it =)

PicoCentauri:
No worries. Thanks to everybody for the work!

PR #120, opened 2022-10-19
Use tox for building docs

PicoCentauri:
This simplifies the local building process for a developer. It especially prevents that if another version of
rascaline is installed the API docs are not updated because sphinx prefers the installed version.


tox.ini
```diff
@@ -35,38 +45,35 @@ commands =
...
-    discover
-
+deps = {[testenv]deps}
 commands =
     python setup.py --quiet install
     discover -p "*.py" -s python/tests
 
+[testenv:docs]
+allowlist_externals =
+    sphinx-build
```
Luthaf, line 64: This should not be needed, tox should be using sphinx-build from its own virtualenv

PicoCentauri:
Before we merge I will update the docs.

PR #123, opened 2022-10-24
Lode center contribution and Madelung tests

PicoCentauri:
Adding a `center_atom_weight` parameter similarly to SOAP.

With this we can also add the Madelung tests for LODE. However with the current set of SphericalExpansionParameters I ran into `missing block` panicks.

```
thread 'calculators::lode::spherical_expansion::tests::madelung' panicked at 'missing block', rascaline/src/calculators/lode/spherical_expansion.rs:580:28
```

This seems to related to the small `atomic_gaussian_width` and `cutoff`. 

TODO
------
- [x] Add center contribution contribution for exponents different then 1

### DIFF
--- a/rascaline/src/calculators/lode/spherical_expansion.rs
+++ b/rascaline/src/calculators/lode/spherical_expansion.rs
@@ -10,16 +10,17 @@ use crate::{Error, System, Vector3D};
 use crate::systems::UnitCell;
 
 use crate::labels::{SamplesBuilder, SpeciesFilter, LongRangePerAtom};
-use crate::labels::{KeysBuilder, CenterSingleNeighborsSpeciesKeys};
+use crate::labels::{KeysBuilder, CenterSingleNeighborsSpeciesKeysSystem};
 
 use crate::calculators::CalculatorBase;
 use crate::calculators::radial_integral::RadialIntegral;
+use crate::calculators::soap::{GtoParameters, SoapGtoRadialIntegral};
 
 use super::LodeRadialBasis;
 
 use crate::math::CachedAllocationsSphericalHarmonics;
 use crate::math::{compute_k_vectors, KVector};
-use crate::math::{gamma_ui, gamma};
+use crate::math::{gamma_ui, gamma, hyp2f1};
 
 
 /// Parameters for LODE spherical expansion calculator.
@@ -46,6 +47,10 @@ pub struct LodeSphericalExpansionParameters {
     pub max_angular: usize,
     /// Width of the atom-centered gaussian used to create the atomic density.
     pub atomic_gaussian_width: f64,
+    /// Weight of the center atom contribution to the features.
+    /// If `1.0` the center atom contribution is weighted the same as any other
+    /// contribution.
+    pub center_atom_weight: f64,
     /// Radial basis to use for the radial integral
     pub radial_basis: LodeRadialBasis,
     /// Potential exponent of the decorated atom density. Currently only
@@ -94,6 +99,8 @@ pub struct LodeSphericalExpansion {
     spherical_harmonics: CachedAllocationsSphericalHarmonics,
     /// implementation + cached allocation to compute the radial integral
     radial_integral: CachedAllocationsRadialIntegral,
+    /// implementation + cached allocation to compute the radial integral
+    rspace_radial_integral: SoapGtoRadialIntegral,
     /// Cached allocations for the k_vector => nlm projection coefficients.
     /// The vector contains different l values, and the Array is indexed by
     /// `m, n, k`.
@@ -177,6 +184,13 @@ impl LodeSphericalExpansion {
 
         let radial_integral = CachedAllocationsRadialIntegral::new(&parameters)?;
 
+        let rspace_gto_parameters = GtoParameters{
+            max_radial: parameters.max_radial,
+            max_angular: parameters.max_angular,
+            atomic_gaussian_width: parameters.atomic_gaussian_width,
+            cutoff: parameters.cutoff };
+        let rspace_radial_integral = SoapGtoRadialIntegral::new(rspace_gto_parameters)?;
+
         let spherical_harmonics = CachedAllocationsSphericalHarmonics::new(
             parameters.max_angular,
         );
@@ -190,6 +204,7 @@ impl LodeSphericalExpansion {
             parameters,
             spherical_harmonics,
             radial_integral,
+            rspace_radial_integral,
             k_vector_to_m_n,
         });
     }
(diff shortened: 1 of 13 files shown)


Luthaf:
Right, this is wrong:

https://github.com/Luthaf/rascaline/blob/f389e9048f12e57981b266f826df45e8e537f40d/rascaline/src/calculators/lode/spherical_expansion.rs#L313-L317

`CenterSingleNeighborsSpeciesKeys` will only generate a key `species_1, species_2` if there is a neighbors with species_2 around a center with species_1, which is not what we need for LODE. We will need to change this to use all the species pairs in the system instead.

PicoCentauri:
Ohh yes I will take a look. Thanks!

rascaline/src/labels/keys.rs
```diff
@@ -32,7 +32,38 @@ impl KeysBuilder for CenterSpeciesKeys {
...
+impl KeysBuilder for CenterSingleNeighborsSpeciesKeysSystem {
+    fn keys(&self, systems: &mut [Box<dyn System>]) -> Result<Labels, Error> {
+
+        let mut all_species_pairs = BTreeSet::new();
+        for system in systems {
+            for &species_first in system.species()? {
+                for &species_second in system.species()? {
+                    if species_first == species_second && !self.self_pairs {
+                        continue;
+                    }
```
Luthaf, line 50: This logic is weird, species_first and species_second could be from two different atoms, so why is `self_pairs` controlling that?

rascaline/src/labels/keys.rs
```diff
@@ -32,7 +32,38 @@ impl KeysBuilder for CenterSpeciesKeys {
 }
 
 /// Compute a set of keys with two variables: the central atom species and a
-/// single neighbor atom species.
+/// all neighbor atom species within the whole system.
+pub struct CenterSingleNeighborsSpeciesKeysSystem {
```
Luthaf, line 36: Alternative name suggestion: `AllSpeciesPairsKeys`

rascaline/tests/lode-madelung.rs
```diff
@@ -0,0 +1,103 @@
+use approx::assert_relative_eq;
+use rascaline::calculators::lode::LodeRadialBasis;
+use rascaline::calculators::{LodeSphericalExpansionParameters, CalculatorBase, LodeSphericalExpansion};
+use rascaline::systems::{System, SimpleSystem, UnitCell};
+use rascaline::{Calculator, Matrix3, Vector3D, CalculationOptions};
+
```
Luthaf, line 6: Might be worth adding a small explanation of what's going on here, what's a Madelung constant for people not familliar with it, and how is it related to lode

rascaline/tests/lode-madelung.rs
```diff
@@ -0,0 +1,103 @@
...
+                let options = CalculationOptions {..Default::default()};
+    
+                let descriptor = calculator.compute(&mut crystal.systems, options).unwrap();
+
+                let madelung = (
+                        crystal.charges[0] * descriptor.block_by_id(0).values().data.as_array()[[0, 0, 0]]
+                      + crystal.charges[1] * descriptor.block_by_id(1).values().data.as_array()[[0, 0, 0]])
+                    / -(4.0 * std::f64::consts::PI * cutoff.powf(2.0)).powf(0.75);
+
+                assert_relative_eq!(madelung, crystal.madelung, max_relative=8e-2);
```
Luthaf, line 99: this is quite a high relative error. Can we make it go down by increasing the k_cutoff?

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -46,6 +46,10 @@ pub struct LodeSphericalExpansionParameters {
     pub max_angular: usize,
     /// Width of the atom-centered gaussian used to create the atomic density.
     pub atomic_gaussian_width: f64,
+    /// Weight of the center atom contribution to the features.
+    /// If `1.0` the center atom contribution is weighted the same as any other
+    /// contribution.
```
Luthaf, line 51: Maybe add something about "this is only for the central atom in the central image ..."

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -489,6 +543,34 @@ impl CalculatorBase for LodeSphericalExpansion {
...
+                let values = block.values_mut();
+                let array = values.data.as_array_mut();
+
+                let sample = [system_i.into(), center_i.into()];
+                let sample_i = match values.samples.position(&sample) {
+                    Some(s) => s,
+                    None => continue
+                };
+
+                for (_property_i, [n]) in values.properties.iter_fixed_size().enumerate() {
```
Luthaf, line 567:
```suggestion
                for (property_i, [n]) in values.properties.iter_fixed_size().enumerate() {
```

Luthaf (commented):
Looks good, there is one slightly fishy thing around `CenterSingleNeighborsSpeciesKeysSystem`

rascaline/tests/lode-madelung.rs
PicoCentauri: The reason is that we probe a wide range of `cutoff`s and` atomic_gaussian_width`s. We could bring down the accuracy to `5e-4` if just restrict the test to the smallest `cutoff` and and `atomic_gaussian_width`...

Luthaf: Could we have two tests/two assert_relative_eq? One which would probe with high accuracy, and the ones which are already here. This way we can still catch small errors if they start occurring later.

rascaline/src/labels/keys.rs
PicoCentauri: To be honest. I don't know why I put this their. Should work fine without this condition...

Luthaf: This might be a copy paste from `CenterSingleNeighborsSpeciesKeys` (which has to deal with this case)

PicoCentauri: Might be. It is gone now.

rascaline/tests/lode-madelung.rs
PicoCentauri: Yes we can split.

rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -729,4 +814,33 @@ mod tests {
...
+
+            let mut spherical_expansion = LodeSphericalExpansion::new(LodeSphericalExpansionParameters {
+                cutoff: 5.0,
+                k_cutoff: None,
+                max_radial: 6,
+                max_angular: 2,
+                atomic_gaussian_width: 1.0,
+                center_atom_weight: 1.0,
+                radial_basis: LodeRadialBasis::splined_gto(1e-8),
+                potential_exponent: *p,
```
Luthaf, line 839: FYI, you can remove the need for a dereference here (i.e. use `potential_exponent: p`) by changing the pattern in the for loop to `for (i, &p) in potential_exponents.iter().enumerate() {`. But this is fine as well!

Luthaf (approved):

PR #127, opened 2022-11-08
Render How-to guides using Sphinx galleries

PicoCentauri:
Render the user and developer how-tos using sphinx galleries. All examples use same dataset and hypers as the tutorials.

This change also makes the `check_examples.py` superflous since all examples will be run with sphinx. 

The current solution for the Rust/C/C++ how-tos are so far not perfect but acceptable. The tab for theses examples are currently shown below the python example.


docs/src/devdoc/how-to/profiling.rst
```diff
@@ -1,65 +0,0 @@
...
-
-- ``Calculator::prepare``: building the list of samples/properties that will be in the descriptor
-- ``XXX::compute``: building blocks for the overall calculation
-- ``NeighborsList``: construction of the list of neighbors
-
-.. tabs::
-
-    .. group-tab:: Python
-
-        .. literalinclude:: ../../../../python/examples/profiling.py
```
Luthaf, line 49: You should be able to use ` .. include:: examples/something.rst` to keep everything inside tabs at the top level, and not have to move the other languages examples to python/examples/something.py (which feels very confusing)

PicoCentauri: Yes, I tried this and it works with ignoring warnings. Sphinx-tabs don't want a title but Sphinx-gallery requires a title...

PicoCentauri: Should be fixed with some workarounds now. @Luthaf tell me if you are happy with this solution.

Luthaf: Oh yes, this is very clever! Since the example is no longer copy/pasteable, should we only have `:start-after:`, to include the final links to download the .py/.ipynb files?

PicoCentauri: Good idea, I thought the same but if you do this the download file is for some reason not reachable for the included rst:

```
Warning, treated as error:
/Users/philiploche/repos/rascaline/docs/src/examples/property-selection.rst:342:download file not readable: /Users/philiploche/repos/rascaline/docs/src/how-to/property-selection.py
```

PR #131, opened 2022-11-15
Add a neighbor list calculator

Luthaf:
This allow to reduce rascaline role to a minimum, and only use it to extract data from atomistic systems. All subsequent calculations could then be written using PyTorch/JAX/...

A neighbor list is also useful when constructing message passing models, to know which atoms to include/exclude from the message passing step.

---

Ping @felixmusil, this might be another way of getting a NL for your pytorch models (once we finalize the rascaline <=> pytorch integration).

### DIFF
--- a/rascaline/src/calculators/neighbor_list.rs
+++ b/rascaline/src/calculators/neighbor_list.rs
@@ -1,0 +1,627 @@
+use std::sync::Arc;
+use std::collections::BTreeSet;
+
+use equistore::TensorMap;
+use equistore::{Labels, LabelsBuilder, LabelValue};
+
+use super::CalculatorBase;
+
+use crate::{Error, System};
+
+
+/// This calculator computes the neighbor list for a given spherical cutoff, and
+/// returns the list of distance vectors between all pairs of atoms strictly
+/// inside the cutoff.
+///
+/// Users can request either a "full" neighbor list (including an entry for both
+/// `i - j` pairs and `j - i` pairs) or save memory/computational by only
+/// working with "half" neighbor list (only including one entry for each `i/j`
+/// pair)
+///
+/// Self pairs (pairs between an atom and periodic copy itself) can appear when
+/// the cutoff is larger than the cell under periodic boundary conditions. Self
+/// pairs with a distance of 0 are not included in this calculator, even though
+/// they are required when computing SOAP.
+///
+/// This sample produces a single property (`"distance"`) with three components
+/// (`"pair_direction"`) containing the x, y, and z component of the vector from
+/// the first atom in the pair to the second. In addition to the atom indexes,
+/// the samples also contain a pair index, to be able to distinguish between
+/// multiple pairs between the same atom (if the cutoff is larger than the
+/// cell).
+#[derive(Debug, Clone)]
+#[derive(serde::Deserialize, serde::Serialize, schemars::JsonSchema)]
+pub struct NeighborList {
+    /// Spherical cutoff to use to determine if two atoms are neighbors
+    pub cutoff: f64,
+    /// Should we compute a full neighbor list (each pair appears twice, once as
+    /// `i-j` and once as `j-i`), or a half neighbor list (each pair only
+    /// appears once)
+    pub full_neighbor_list: bool,
+}
+
+/// Sort a pair and return true if the pair was inverted
+fn sort_pair((i, j): (i32, i32)) -> ((i32, i32), bool) {
+    if i <= j {
+        ((i, j), false)
+    } else {
+        ((j, i), true)
+    }
+}
+
+impl CalculatorBase for NeighborList {
+    fn name(&self) -> String {
+        "neighbors list".into()
+    }
+
+    fn parameters(&self) -> String {
+        serde_json::to_string(self).expect("failed to serialize to JSON")
+    }
+
+    fn keys(&self, systems: &mut [Box<dyn System>]) -> Result<Labels, Error> {
+        assert!(self.cutoff > 0.0 && self.cutoff.is_finite());
+
+        if self.full_neighbor_list {
+            FullNeighborList { cutoff: self.cutoff }.keys(systems)
+        } else {
+            HalfNeighborList { cutoff: self.cutoff }.keys(systems)
+        }
+    }
+
+    fn samples_names(&self) -> Vec<&str> {
+        return vec!["structure", "pair_id", "first_atom", "second_atom"];
+    }
+
+    fn samples(&self, keys: &Labels, systems: &mut [Box<dyn System>]) -> Result<Vec<Arc<Labels>>, Error> {
+        assert!(self.cutoff > 0.0 && self.cutoff.is_finite());
+
+        if self.full_neighbor_list {
+            FullNeighborList { cutoff: self.cutoff }.samples(keys, systems)
+        } else {
+            HalfNeighborList { cutoff: self.cutoff }.samples(keys, systems)
+        }
+    }
+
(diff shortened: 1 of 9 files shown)


felixmusil:
Sounds like a very useful feature. For pytorch one might also need to have a C/CPP interface so that it can be used when interfacing with MD codes.

Luthaf:
rascaline has a C and C++ interface (https://github.com/Luthaf/rascaline/tree/master/rascaline-c-api/include, https://luthaf.fr/rascaline/latest/references/api/cxx/index.html), which are used in the (prototype) torch bindings (integrating any rascaline calculator, including this one with torch) here: https://github.com/Luthaf/rascaline-torch.

rascaline/src/calculators/neighbor_list.rs
```diff
@@ -0,0 +1,620 @@
...
+
+use equistore::TensorMap;
+use equistore::{Labels, LabelsBuilder, LabelValue};
+
+use super::CalculatorBase;
+
+use crate::{Error, System};
+
+
+/// This calculator compute the neighbor list for a given spherical cutoff, and
```
PicoCentauri, line 12:
```suggestion
/// This calculator computes the neighbor list for a given spherical cutoff, and
```

rascaline/src/calculators/neighbor_list.rs
```diff
@@ -0,0 +1,620 @@
...
+/// returns the list of distance vectors between pairs of atoms strictly inside
```
PicoCentauri, line 13:
```suggestion
/// returns the list of distance vectors between all pairs of atoms strictly inside
```

rascaline/src/calculators/neighbor_list.rs
```diff
@@ -0,0 +1,620 @@
...
+
+    fn parameters(&self) -> String {
+        serde_json::to_string(self).expect("failed to serialize to JSON")
+    }
+
+    fn keys(&self, systems: &mut [Box<dyn System>]) -> Result<Labels, Error> {
+        assert!(self.cutoff > 0.0 && self.cutoff.is_finite());
+
+        if self.full_neighbor_list {
+            FullNeighborList { cutoff: self.cutoff }.keys(systems)
```
PicoCentauri, line 56: You probably considered this but are the implementations of `FullNeighborList` and `HalfNeighborList` really so different to not distinguish the cases within one implementation?

From an naive point of view you could just calculate half the neighbor list and stack them with inverted species.

rascaline/src/calculators/neighbor_list.rs
```diff
@@ -0,0 +1,620 @@
...
+        return Ok(results);
+    }
+
+    fn compute(&mut self, systems: &mut [Box<dyn System>], descriptor: &mut TensorMap) -> Result<(), Error> {
+        for (system_i, system) in systems.iter_mut().enumerate() {
+            system.compute_neighbors(self.cutoff)?;
+            let species = system.species()?;
+
+            for (pair_id, pair) in system.pairs()?.iter().enumerate() {
+                // TODO explain
```
PicoCentauri, line 196: Yes an explanation would be nice here.

rascaline/src/calculators/neighbor_list.rs
```diff
@@ -0,0 +1,620 @@
...
+}
+
+/// Sort a pair and return true if the pair was inverted
+fn sort_pair((i, j): (i32, i32)) -> ((i32, i32), bool) {
+    if i <= j {
+        ((i, j), false)
+    } else {
+        ((j, i), true)
+    }
+}
```
PicoCentauri, line 261: maybe for clarity move this to the top of the file.

rascaline/src/calculators/neighbor_list.rs
```diff
@@ -0,0 +1,620 @@
...
+
+        let array = block.data.as_array();
+        let expected = &ndarray::arr3(&[
+            [[0.0], [-1.5109], [0.0]]
+        ]).into_dyn();
+        assert_relative_eq!(array, expected, max_relative=1e-6);
+    }
+
+    #[test]
+    fn full_neighbor_list() {
```
PicoCentauri, line 495: I don't really see here where you test the `full_neighbor_list`. The test looks exactly the same as the `half_neighbor_list`.

PicoCentauri (changes requested):
Overall really good. Especially, the tricks in the `HalfNeighborList` is good. I have some points about clarity in the code, the tests and about code sharing between the two implementations.

PicoCentauri:
Also a description of the new calculator in the README.rst would be nice.

rascaline/src/calculators/neighbor_list.rs
Luthaf: the `full_neighbor_list` hyper is true, and there are more pairs in this test. but yes, overall the structure is very similar

rascaline/src/calculators/neighbor_list.rs
Luthaf: Given that three functions need to change depending on the half/full distinction, I feel that having separate structs/impl makes it clearer what's going on.

The issue with doing a half NL and then duplicating, is that we end up doing quite a bit of redundant work (sorting species for the half, only to unsort them later), for a calculator that should be a small wrapper on top of the nl and thus should get out of the way as fast as possible.

PicoCentauri: I see.

rascaline/src/calculators/neighbor_list.rs
PicoCentauri: Okay, maybe I just oversaw the changes this morning.

PicoCentauri (approved):
Thanks for the changes. LGTM

felixmusil:
I could not understand from the PR changes what would be the features produced by the NL computation so I missed that it is missing a cell shift index `S`.

In short, this is needed so that 
```
nl = NeighborList(cutoff, True)
nl_mapping = nl.compute(frame)
``` 
match the output of `ase.neighborlist.neighbor_list('ijS', frame, cutoff)`. This is needed to compute direction vectors with periodic boundary conditions in auto-differentiation frameworks. Also, it allows the computation of virals in a simple manner.

Such shift index array seems to already be a byproduct of the `NeighborList` computation. How difficult it would be to add this additional information to `nl_mapping` ?

felixmusil:
> S can be recovered if needed by doing S = r_ij - r_j + r_i, but I found that it is not necessary for the calculation of virials (see around 

Indeed it's possible but it seems very natural...

> Integration with auto-diff is provided in https://github.com/Luthaf/rascaline-torch for arbitrary representation, soon including the neighbor list as well. You can also do the same thing manually by writing a custom Autograd function (in Torch terms), where the backward pass does the Jacobian-vector product using the forward gradients produced by rascaline. For the neighbors list, the gradients w.r.t. positions are only -1/0/1, so I don't see where you need the S?

That's not the kind of integration I am referring to. What you suggest is too much work compared to just recomputing the distances and direction vectors within the auto-diff framework.

> Finally, if there is a stronger need for it, we could add an option to also provide and store the cell shift indices within this calculator, it should be relatively easy.

That would be ideal to make this NL useful in many scenarios.

Luthaf:
> That's not the kind of integration I am referring to. What you suggest is too much work compared to just recomputing the distances and direction vectors within the auto-diff framework.

Sure, but if you only want to get the S vectors and do not care about using rascaline/equistore infrastructure, what do you gain by using this instead of the ASE neighbor list? I'm trying to understand your use case.

Luthaf:
Ok, that makes sense =) I'll give this a go soon-ish!

PR #132, opened 2022-11-16
Refactor SphericalHarmonicsCache and RadialIntegralCache out of the spherical expansion code

Luthaf:
These will also be used by LODE and the spherical expansion by pair


PicoCentauri:
I like moving stuff in math.

PR #133, opened 2022-11-17
Fix potential exponents >1 for LODE

PicoCentauri:
Fix even potential exponents `[2, 3, 4, 5, 6]` for LODE using an analytical expression for the gamma function over a powerlaw as @kvhuguenin is using them in pyLODE.

### DIFF
--- a/rascaline/src/calculators/lode/spherical_expansion.rs
+++ b/rascaline/src/calculators/lode/spherical_expansion.rs
@@ -19,7 +19,7 @@ use super::LodeRadialBasis;
 
 use crate::math::CachedAllocationsSphericalHarmonics;
 use crate::math::{compute_k_vectors, KVector};
-use crate::math::{gamma_ui, gamma, hyp2f1};
+use crate::math::{expi, erfc, gamma, hyp2f1};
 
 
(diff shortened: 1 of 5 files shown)


rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -257,14 +257,33 @@ impl LodeSphericalExpansion {
...
 
             for k_vector in k_vectors {
                 let k_norm_squared = k_vector.norm * k_vector.norm;
-                let value = gamma_ui(0.5 * p_eff, 0.5 * k_norm_squared * smearing_squared);
-                fourrier.push(factor * value / k_vector.norm.powi(p_eff as i32));
+                let x = 0.5 * k_norm_squared * smearing_squared;
+
+                // Compute the gamma_ui over a power law using analytical
+                // expressions for a more stable Fourier transform of the
+                // density
```
Luthaf, line 268: @kvhuguenin what's the exact function being computed here? would be nice to add it to the comment

Luthaf:
Close & open to restart CI (it was not running for some reason)

PR #134, opened 2022-11-21
Added selected_keys option to the compute function to allow manual selection of blocks for calculation

hurricane642:
Hi all! This PR has added a new option `selected_keys` to the `compute` function. It allows the user to manually select the keys whose blocks will be computed by the algorithm. All relevant tests have been added.

### DIFF
--- a/python/tests/calculators.py
+++ b/python/tests/calculators.py
@@ -419,6 +419,172 @@ class TestComputePartialProperties(unittest.TestCase):
         )
 
 
+class TestComputeSelectedKeys(unittest.TestCase):
+    def test_selection_existing(self):
+        system = TestSystem()
+        calculator = DummyCalculator(cutoff=3.2, delta=2, name="")
+
+        # Manually select the keys
+        selected_keys = Labels(
+            names=["species_center"],
+            values=np.array([[1]], dtype=np.int32),
+        )
+        descriptor = calculator.compute(
+            system, use_native_system=False, selected_keys=selected_keys
+        )
+
+        self.assertEqual(len(descriptor.keys), 1)
+        self.assertEqual(tuple(descriptor.keys[0]), (1,))
+
+    def test_select_key_not_in_systems(self):
+        system = TestSystem()
+        calculator = DummyCalculator(cutoff=3.2, delta=2, name="")
+
+        # Manually select the keys
+        selected_keys = Labels(
+            names=["species_center"],
+            values=np.array([[4]], dtype=np.int32),
+        )
+        descriptor = calculator.compute(
+            system, use_native_system=False, selected_keys=selected_keys
+        )
+
+        C_block = descriptor.block(species_center=4)
+        self.assertEqual(C_block.values.shape, (0, 2))
+
+    def test_predefined_selection(self):
+        system = TestSystem()
+        calculator = DummyCalculator(cutoff=3.2, delta=2, name="")
+
+        selected_keys = Labels(
+            names=["species_center"],
+            values=np.array([[1]], dtype=np.int32),
+        )
+
+        keys = Labels(
+            names=["species_center"],
+            values=np.array([[1], [8]], dtype=np.int32),
+        )
+
+        # selection from TensorMap
+        selected = [
+            Labels(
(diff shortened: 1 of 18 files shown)


python/tests/calculators.py
```diff
@@ -419,6 +419,172 @@ def test_errors(self):
...
+            values=np.array([[1]], dtype=np.int32),
+        )
+        descriptor = calculator.compute(
+            system, use_native_system=False, selected_keys=selected_keys
+        )
+
+        self.assertEqual(len(descriptor.keys), 1)
+        self.assertEqual(tuple(descriptor.keys[0]), (1,))
+
+    def test_select_defunct(self):
```
Luthaf, line 439:
```suggestion
    def test_select_key_not_in_systems(self):
```

rascaline-c-api/src/calculator.rs
```diff
@@ -251,6 +265,8 @@ pub struct rascal_calculation_options_t {
     selected_samples: rascal_labels_selection_t,
     /// Selection of properties to compute for the samples
     selected_properties: rascal_labels_selection_t,
+    /// Keys which will be in the resulting representation
```
Luthaf, line 268:
```suggestion
    /// Selection for the keys to include in the output. Set this parameter 
    /// to `NULL` to use the default set of keys, as determined by the calculator.
    /// Note that this default set of keys can depend on which systems 
    /// we are running the calculation
```

rascaline/src/calculator.rs
```diff
@@ -185,6 +194,8 @@ pub struct CalculationOptions<'a> {
     pub selected_samples: LabelsSelection<'a>,
     /// Selection of properties to compute for the samples
     pub selected_properties: LabelsSelection<'a>,
+    /// Keys which will be in the resulting representation
```
Luthaf, line 197:
```suggestion
    /// Selection for the keys to include in the output. If this is `None`, the 
    /// default set of keys (as determined by the calculator) will be used. 
    /// Note that this default set of keys can depend on which systems 
    /// we are running the calculation
```

rascaline/src/calculator.rs
```diff
@@ -245,15 +257,30 @@ impl Calculator {
...
     }
 
-    /// Get the set of keys this calculator would produce for the given systems
-    pub fn default_keys(&self, systems: &mut [Box<dyn System>]) -> Result<Labels, Error> {
-        self.implementation.keys(systems)
-    }
 
     #[time_graph::instrument(name="Calculator::prepare")]
     fn prepare(&mut self, systems: &mut [Box<dyn System>], options: CalculationOptions,) -> Result<TensorMap, Error> {
         // TODO: allow selecting a subset of keys?
```
Luthaf, line 263: You can now remove this TODO!

rascaline/src/calculator.rs
```diff
@@ -245,15 +257,30 @@ impl Calculator {
...
+                            default_keys.names().join(", "),
+                            keys.names().join(", "))
+                            ));
+                    }
+                    else {
+                        return Err(Error::InvalidParameter(format!("Empty selected keys")));
+                    }
+                }
+            },
+            None => default_keys,
```
Luthaf, line 282: This can be made a bit clearer with 

```suggestion
            Some(keys) if keys.is_empty() => {
                return Err(Error::InvalidParameter(format!("Empty selected keys")));
            }
            Some(keys) => {
                if default_keys.names() == keys.names() {
                    keys.clone()
                } else {
                    return Err(Error::InvalidParameter(format!(
                        "The key names of the calculator [{}] and the passed keys [{}] do not match:",
                        default_keys.names().join(", "),
                        keys.names().join(", "))
                    ));
                }
            }
            None => default_keys,
```

Luthaf:
@hurricane642 does everything makes sense to you here? Should we also add an example in the same spirit as the ones for sample selection & properties selection?

hurricane642:
> @hurricane642 does everything makes sense to you here? Should we also add an example in the same spirit as the ones for sample selection & properties selection?

Yes, everything is fine. I think doing a how-to would be helpful.

PR #137, opened 2022-11-24
Transformer class for the property selection

hurricane642:
Hello, everyone!
This PR is the first step in adding the 'Transformer' class to rascaline. The 'Transformer' class makes it easy to create a representation matrix when using some other matrix as a reference. A classic use case is to create a TensorMap representation for a dataset, then perform transformations within that TensorMap (e.g., keys_to_features or keys_to_properties), and select the most useful features in the transformed TensorMap. The 'Transformer' allows a set of these features to be used to calculate a new TensorMap, thus saving computation time and maintaining a single I tried to explain the logic of the algorithm in the comments to the code, but if you have any questions - ask, I'll be happy to answer, as well as happy to receive any comments and suggestions.

IMPORTANT:
We need to add to the code the ability to perform several transformations one by one. This will be the next step.
Note that this branch is based on the selected_keys branch, which itself is still a PR. representation for all representations.

### DIFF
--- a/python/rascaline/transformer.py
+++ b/python/rascaline/transformer.py
@@ -1,0 +1,296 @@
+# -*- coding: utf-8 -*-
+import numpy as np
+from equistore import Labels, TensorBlock, TensorMap
+
+
+class Transformer:
+    """The 'Transformer' class makes it easy to create a representation matrix
+    when using some other matrix as a reference. A classic use case is to create
+    a TensorMap representation for a dataset, then perform transformations
+    within that TensorMap (e.g., keys_to_features or keys_to_properties), and
+    select the most useful features in the transformed TensorMap.
+    The 'Transformer' allows a set of these features to be used to calculate
+    a new TensorMap, thus saving computation time and maintaining a single
+    representation for all representations.
+
+    Parameters:
+    -----------
+    :param selectors: The selector to be used when selecting the features.
+        Currently the same selector is used for different blocks.
+        #TODO: It should be possible for the user to pass a list of selectors
+        for blocks.
+    :param transformation: the type of transformation to be performed.
+        Two options are possible - 'keys_to_features' and 'keys_to_properties'.
+        #TODO: provide the ability to pass a list of conversions that will occur
+        one after the other.
+    :param moved_keys: Those keys which will be moved during the transformation.
+        This variable can accept knowledge of type str (one key), list (a list
+        of keys) and Labels (in addition to the name of the keys, pass a list
+        of keys to be moved).
+    """
+
+    def __init__(self, selector, transformation=None, moved_keys=None):
+        #
+        self.selector = selector
+        self.transformation = transformation
+        self.moved_keys = moved_keys
+        if (
+            (self.transformation is not None)
+            and (self.transformation != "keys_to_samples")
+            and (self.transformation != "keys_to_properties")
+        ):
+            raise ValueError(
+                "`transformation` parameter should be either `keys_to_samples`,"
+                f" either `keys_to_properties`, got {self.transformation}"
+            )
+        if (self.transformation is None) and (self.moved_keys is not None):
+            raise ValueError("unable to shift keys: unknown transformation type")
+
+    def _copy(self, tensor_map):
+        """This function that allows you to create a copy of 'TensorMap'.
+        It may be worth adding this function to the equistore.
+        """
+        blocks = []
+        for _, block in tensor_map:
+            blocks.append(block.copy())
+        return TensorMap(tensor_map.keys, blocks)
+
+    def keys_definition(self):
+        """This is another internal function that performs two main tasks.
+        First, it converts all moved_keys to the same format.  What is
+        meant is that further we need the names of the keys we are going
+        to move, as well as the 'TensorMap' keys, which will be passed
+        to the compute function as a reference at the end. This function
+        stores the names of the moved keys in the 'moved_keys_names' array,
+        and stores the keys of the final TensorMap reference in 'final_keys'.
+        """
+        # the first 2 cases are simple - we either copy the moved_keys directly,
+        # or create an array based on them, and simply take all the keys passed
+        # in the fit TensorMap step as the final keys.
+        if isinstance(self.moved_keys, str):
+            self.moved_keys_names = [self.moved_keys.copy()]
+            self.final_keys = self._old_keys
+        elif isinstance(self.moved_keys, list):
+            self.moved_keys_names = self.moved_keys.copy()
+            self.final_keys = self._old_keys
+        else:
+            # The third case is a little more complicated.
+            # First, we save the names of the moved keys,
+            # taking them from Labels 'moved_keys'.
+            self.moved_keys_names = self.moved_keys.names
+            names = []
+            new_keys = []
+            # Let's write down the order of the keys we will have during the
+            # course of the algorithm in the 'names'
+            names.extend(self.tensor_map.keys.names)
+            names.extend(self.moved_keys_names)
+            # Now let's generate reference TensorMap keys. They will consist of
+            # two parts - those keys that were left after transformation, and
+            # those keys that were in the values of the variable moved_keys.
+            # Go through them and create all possible combinations of these
+            # parts.
+            for key in self.tensor_map.keys:
+                for value in self.moved_keys:
+                    clue = [k.copy() for k in key]
+                    clue.extend(value)
+                    new_keys.append(clue)
+            # The keys have been listed in random order, let's arrange them and
+            # store the values in 'final_keys'.
+            indices = []
+            for key in self._old_keys_names:
+                indices.append(names.index(key))
+            ordered_keys = []
+            for el in new_keys:
+                key = [el[i] for i in indices]
+                ordered_keys.append(key)
+            self.final_keys = Labels(
+                names=self._old_keys_names, values=np.array(ordered_keys)
+            )
+
+    def _mover(self, tensor_map):
+        # Internal function that does the transformation of the reference
+        # Tensormap.
+        self._old_keys = tensor_map.keys
+        self._old_keys_names = tensor_map.keys.names
+        tensor_copy = self._copy(tensor_map)
+        if self.transformation is not None:
+            if self.transformation == "keys_to_samples":
+                tensor_copy.keys_to_samples(self.moved_keys)
+            elif self.transformation == "keys_to_properties":
+                tensor_copy.keys_to_properties(self.moved_keys)
+        return tensor_copy
+
+    def properties_selection(self):
+        # This function selects properties according to a preset algorithm
+        # within each 'TensorMap' block
+        blocks = []
+        for _, block in self.tensor_map:
+            mask = self.selector.fit(block.values).get_support()
+            selected_properties = block.properties[mask]
+            blocks.append(
+                TensorBlock(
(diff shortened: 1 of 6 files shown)


python/rascaline/transformer.py
```diff
@@ -0,0 +1,296 @@
+# -*- coding: utf-8 -*-
+import numpy as np
+from equistore import Labels, TensorBlock, TensorMap
+
+
+class Transformer:
```
Luthaf, line 6: how about something like this? Then explaining in the docstring why it is useful.

```suggestion
class PropertiesSelector:
```

python/rascaline/transformer.py
```diff
@@ -0,0 +1,296 @@
...
+        #TODO: It should be possible for the user to pass a list of selectors
+        for blocks.
+    :param transformation: the type of transformation to be performed.
+        Two options are possible - 'keys_to_features' and 'keys_to_properties'.
+        #TODO: provide the ability to pass a list of conversions that will occur
+        one after the other.
+    :param moved_keys: Those keys which will be moved during the transformation.
+        This variable can accept knowledge of type str (one key), list (a list
+        of keys) and Labels (in addition to the name of the keys, pass a list
+        of keys to be moved).
```
Luthaf, line 29:
```suggestion
        This variable can be anything supported by the :py:class:`equistore.TensorMap.keys_to_properties`
        or :py:class:`equistore.TensorMap.keys_to_samples` functions, i.e. one string, a list of strings or an 
        instance of :py:class:`equistore.Labels`
```

python/rascaline/transformer.py
```diff
@@ -0,0 +1,296 @@
...
+    """
+
+    def __init__(self, selector, transformation=None, moved_keys=None):
+        #
+        self.selector = selector
+        self.transformation = transformation
+        self.moved_keys = moved_keys
```
Luthaf, line 36: The user should not access these after creating an instance of this class

```suggestion
        self._selector = selector
        self._transformation = transformation
        self._moved_keys = moved_keys
```

python/rascaline/transformer.py
```diff
@@ -0,0 +1,296 @@
...
+            and (self.transformation != "keys_to_properties")
+        ):
+            raise ValueError(
+                "`transformation` parameter should be either `keys_to_samples`,"
+                f" either `keys_to_properties`, got {self.transformation}"
+            )
+        if (self.transformation is None) and (self.moved_keys is not None):
+            raise ValueError("unable to shift keys: unknown transformation type")
+
+    def _copy(self, tensor_map):
```
Luthaf, line 49: why do you need to make a copy?

python/rascaline/transformer.py
```diff
@@ -0,0 +1,296 @@
...
+        """This function that allows you to create a copy of 'TensorMap'.
+        It may be worth adding this function to the equistore.
+        """
+        blocks = []
+        for _, block in tensor_map:
+            blocks.append(block.copy())
+        return TensorMap(tensor_map.keys, blocks)
+
+    def keys_definition(self):
```
Luthaf, line 58:
```suggestion
    def _keys_definition(self):
```

python/rascaline/transformer.py
```diff
@@ -0,0 +1,296 @@
...
+        to move, as well as the 'TensorMap' keys, which will be passed
+        to the compute function as a reference at the end. This function
+        stores the names of the moved keys in the 'moved_keys_names' array,
+        and stores the keys of the final TensorMap reference in 'final_keys'.
+        """
+        # the first 2 cases are simple - we either copy the moved_keys directly,
+        # or create an array based on them, and simply take all the keys passed
+        # in the fit TensorMap step as the final keys.
+        if isinstance(self.moved_keys, str):
+            self.moved_keys_names = [self.moved_keys.copy()]
```
Luthaf, line 71: copy is not needed, python strings are immutable

```suggestion
            self.moved_keys_names = [self.moved_keys]
```

python/rascaline/transformer.py
```diff
@@ -0,0 +1,296 @@
...
+            self.final_keys = self._old_keys
+        elif isinstance(self.moved_keys, list):
+            self.moved_keys_names = self.moved_keys.copy()
+            self.final_keys = self._old_keys
```
Luthaf, line 75: `final_keys` and `old_keys` are not very clear names. What do they refer to?

python/rascaline/transformer.py
```diff
@@ -0,0 +1,296 @@
...
+            (self.transformation is not None)
+            and (self.transformation != "keys_to_samples")
+            and (self.transformation != "keys_to_properties")
+        ):
+            raise ValueError(
+                "`transformation` parameter should be either `keys_to_samples`,"
+                f" either `keys_to_properties`, got {self.transformation}"
+            )
+        if (self.transformation is None) and (self.moved_keys is not None):
+            raise ValueError("unable to shift keys: unknown transformation type")
```
Luthaf, line 47: I would initialize all attributes here (`moved_keys_names`, `final_keys`, etc.) to `None`, mainly to have a single place where they are all defined. Right now it would be possible to call functions in the wrong order and get attribute undefined errors

python/rascaline/transformer.py
```diff
@@ -0,0 +1,296 @@
...
+        # or create an array based on them, and simply take all the keys passed
+        # in the fit TensorMap step as the final keys.
+        if isinstance(self.moved_keys, str):
+            self.moved_keys_names = [self.moved_keys.copy()]
+            self.final_keys = self._old_keys
+        elif isinstance(self.moved_keys, list):
+            self.moved_keys_names = self.moved_keys.copy()
+            self.final_keys = self._old_keys
+        else:
+            # The third case is a little more complicated.
```
Luthaf, line 77:
```suggestion
            assert isinstance(self.moved_keys, Labels)

            # The third case is a little more complicated.
```

python/rascaline/transformer.py
```diff
@@ -0,0 +1,296 @@
...
+        Two options are possible - 'keys_to_features' and 'keys_to_properties'.
+        #TODO: provide the ability to pass a list of conversions that will occur
+        one after the other.
+    :param moved_keys: Those keys which will be moved during the transformation.
+        This variable can accept knowledge of type str (one key), list (a list
+        of keys) and Labels (in addition to the name of the keys, pass a list
+        of keys to be moved).
+    """
+
+    def __init__(self, selector, transformation=None, moved_keys=None):
```
Luthaf, line 32: we should use the same name as the equistore functions

```suggestion
    def __init__(self, selector, transformation=None, keys_to_move=None):
```

python/tests/transformer.py
```diff
@@ -0,0 +1,127 @@
+# -*- coding: utf-8 -*-
+import unittest
+
+import numpy as np
+from equistore import TensorBlock, TensorMap
+from skcosmo.feature_selection import FPS as FPS_f
```
Luthaf, line 6: why do you rename FPS to FPS_f?

python/rascaline/transformer.py
```diff
@@ -0,0 +1,296 @@
...
+        Parameters:
+        -----------
+        :param tensor_map: reference TensorMap, with which transformations are
+            carried out and in which properties are selected.
+        """
+        self.tensor_map = self._mover(tensor_map)
+        self.keys_definition()
+        self.properties_selection()
+
+    def transform(self, frames, calculator):
```
Luthaf, line 157: Should this class take the calculator in the constructor? Given that it will only be usable with one calculator.

python/rascaline/transformer.py
```diff
@@ -0,0 +1,296 @@
...
+        self._old_keys_names = tensor_map.keys.names
+        tensor_copy = self._copy(tensor_map)
+        if self.transformation is not None:
+            if self.transformation == "keys_to_samples":
+                tensor_copy.keys_to_samples(self.moved_keys)
+            elif self.transformation == "keys_to_properties":
+                tensor_copy.keys_to_properties(self.moved_keys)
+        return tensor_copy
+
+    def properties_selection(self):
```
Luthaf, line 123:
```suggestion
    def _properties_selection(self):
```

python/rascaline/transformer.py
Luthaf: And this function also needs to handle all other kwargs that could be given to the calculator (mainly `gradients` & `use_native_system`). We could either have them explicitly, or take a `**kwargs` argument, set the corresponding `selected_properties` and pass it to the calculator

Luthaf (commented):
the code is not very clear to me, so let's try to iterate a bit on the attributes/functions naming 😃

python/rascaline/transformer.py
hurricane642: Yep, sounds great!

python/rascaline/transformer.py
hurricane642: In the `_mover` function, I move keys_to_samples/keys_to_properties with the tensor, saving its keys before doing so. Before I used the `_copy` function, the saved keys started pointing to the keys of the already modified tensor.

python/tests/transformer.py
hurricane642: I forgot to rename it from the time I thought it would be necessary to make a sample selection too. I'll rename it

python/rascaline/transformer.py
hurricane642: On the first point, it's an interesting question. So, it is clear that we need to input `frames` (our general task is to get `descriptor` from them as simply as possible). For this, we need a `calculator` one way or another. And now the question is at what point we will feed it as input? We either have to do it in the `fit` function or already in `transform`. In general, the first option is possible (we can pass either `descriptor` and `calculator` or `frames` and `calculator` to `fit`). What do you think would be better?
On the second point, yes, I agree, I'll try to fix that.

python/rascaline/transformer.py
Luthaf: Right. If you only need the keys, it might be better to only copy the keys, they should work fine with `tensor.keys.copy()`

python/rascaline/transformer.py
Luthaf: how about `initial_keys`/`transformed_keys`? Since `keys_to_properties` is a "transformation"

python/rascaline/transformer.py
Luthaf: I would pass the calculator even earlier (in `__init__`), and then both `fit` and `transform` take a set of frames + some options as kwargs

hurricane642:
> You can also add skcosmo to the tox environments to be able to run the tests.

Ok, I added it, but it's still having trouble passing the tests. Is there anything else I should add?

Luthaf:
You can skip tests if skcosmo is not installed (as long as it is still tested in the `all-deps` environement), see for example https://github.com/Luthaf/rascaline/blob/3a28ad0831f6f0bf6cb81defd26a47a0e25f228a/python/tests/systems/chemfiles.py#L17

setup.cfg
```diff
@@ -21,6 +21,7 @@ package_dir =
 install_requires =
     numpy
     equistore @ https://github.com/lab-cosmo/equistore/archive/6ca7fa3.zip
+    skcosmo
```
Luthaf, line 24: This should not be in the `install_requires` (it is an optional dependency, like ase or chemfiles)

Luthaf:
Closing for now, we can revisit this later if we still need it!

PR #139, opened 2022-12-01
Adding a conversion function for other packages

rosecers:
Just because I was writing it for myself and figured others would fine it useful. Added all of the necessary tests and included a shout-out in docs where I thought it was appropriate.

### DIFF
--- a/python/rascaline/utils.py
+++ b/python/rascaline/utils.py
@@ -4,6 +4,8 @@ import ctypes
 from ._c_api import RASCAL_BUFFER_SIZE_ERROR
 from .status import RascalError
 
+import warnings
+
 
 def _call_with_growing_buffer(callback, initial=1024):
     bufflen = initial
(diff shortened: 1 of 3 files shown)


python/rascaline/utils.py
```diff
@@ -20,3 +22,225 @@ def _call_with_growing_buffer(callback, initial=1024):
             else:
                 raise
     return buffer.value.decode("utf8")
+
+
+def convert_old_hyperparameter_names(hyperparameters, mode="librascal"):
```
Luthaf, line 27: I would not put a default here, instead either guess from the hypers or force the user to specify what they want

```suggestion
def convert_old_hyperparameter_names(hyperparameters, mode):
```

python/rascaline/utils.py
```diff
@@ -20,3 +22,225 @@ def _call_with_growing_buffer(callback, initial=1024):
...
+            warnings.warn(
+                "{} are not required parameters in the rascaline software infrastructure".format(
+                    ",".join(
+                        [f"`{d}`" for d in deprecated_params if d in hyperparameters]
+                    )
+                )
+            )
+
+        not_supported = [
+            "coefficient_subselection",
```
Luthaf, line 167: `coefficient_subselection` could point people to the `selected_properties` argument of the `compute` function

python/rascaline/utils.py
```diff
@@ -20,3 +22,225 @@ def _call_with_growing_buffer(callback, initial=1024):
...
+                        "rate": params.get("rate", None),
+                        "scale": params.get("scale", None),
+                    }
+                }
+
+        deprecated_params = [
+            "global_species",
+            "expansion_by_species_method",
+            "soap_type",
+            "compute_gradients",
```
Luthaf, line 155: `compute_gradients` could point people to the `gradients` argument of the `compute` function

python/rascaline/utils.py
```diff
@@ -20,3 +22,225 @@ def _call_with_growing_buffer(callback, initial=1024):
...
+                    ",".join(
+                        [f"`{d}`" for d in deprecated_params if d in hyperparameters]
+                    )
+                )
+            )
+
+        not_supported = [
+            "coefficient_subselection",
+            "covariant_lambda",
+            "gaussian_sigma_type",
```
Luthaf, line 169: `gaussian_sigma_type="Constant"` (which is the only one implemented by librascal) is also what rascaline is doing by default

python/rascaline/utils.py
```diff
@@ -20,3 +22,225 @@ def _call_with_growing_buffer(callback, initial=1024):
...
+            "inversion_symmetry",
+            "normalize",
+            "optimization_args",
+            "optimization",
```
Luthaf, line 173: Some of these could be translated to "Gto" radial basis options in rascaline

python/rascaline/utils.py
```diff
@@ -20,3 +22,225 @@ def _call_with_growing_buffer(callback, initial=1024):
...
+                            key
+                            for key in hyperparameters
+                            if key not in anticipated_hypers
+                        ]
+                    )
+                )
+            )
+
+        new_hypers["atomic_gaussian_width"] = hyperparameters.pop(
+            "gaussian_sigma_constant", None
```
Luthaf, line 122: librascal has a lot of default values: `gaussian_sigma_constant` defaults to 0.3, `cutoff_smooth_width` defaults to 0.5, …

Should this code use the librascal defaults when translating to rascaline?

python/rascaline/utils.py
Luthaf, line 27:
```suggestion
def translate_soap_hyperparameters(hyperparameters, mode="librascal"):
```

since this is soap-specific

python/rascaline/utils.py
```diff
@@ -20,3 +22,225 @@ def _call_with_growing_buffer(callback, initial=1024):
...
+            - `rbf`
+            - `rcut`
+            - `sigma`
+            - `sparse`
+            - `species`
+
+    """
+    new_hypers = {}
+
+    if mode == "librascal":
```
Luthaf, line 85: for readability & extensibility, I would have separate functions for each mode: `_translate_soap_hyperparameters_from_librascal`, `_translate_soap_hyperparameters_from_dscribe` and potentially later `_translate_soap_hyperparameters_from_quip`

python/rascaline/utils.py
```diff
@@ -20,3 +22,225 @@ def _call_with_growing_buffer(callback, initial=1024):
...
+        new_hypers["max_angular"] = hyperparameters.pop("lmax", None)
+        new_hypers["max_radial"] = hyperparameters.pop("nmax", None)
+        new_hypers["cutoff"] = hyperparameters.pop("rcut", None)
+
+        if  'rbf' in hyperparameters:
+            new_hypers["radial_basis"] = {hyperparameters.pop("rbf").title(): {}}
+            if new_hypers["radial_basis"] != "Gto":
+                warnings.warn("WARNING: rascaline currently only supports a Gto basis.")
+
+        deprecated_params = ["average", "sparse", "dtype"]
```
Luthaf, line 221: `average="outer"` could point to `equistore.operations.mean_over_samples`, and `average="off"` is the default

python/rascaline/utils.py
```diff
@@ -20,3 +22,225 @@ def _call_with_growing_buffer(callback, initial=1024):
...
+            "nmax",
+            "lmax",
+            "species",
+            "sigma",
+            "rbf",
+            "periodic",
+            "crossover",
+            "average",
+            "sparse",
+            "dtype",
```
Luthaf, line 195: there is also `weighting` which can be translated to either `radial_scaling` (`weighting/function/pow` in dscribe) or `central_atom_weight` (`weighting/w0` in dscribe)

python/rascaline/utils.py
```diff
@@ -20,3 +22,225 @@ def _call_with_growing_buffer(callback, initial=1024):
...
+                "{} are not required parameters in the rascaline software infrastructure".format(
+                    ",".join(
+                        [f"`{d}`" for d in deprecated_params if d in hyperparameters]
+                    )
+                )
+            )
+
+        not_supported = [
+            "periodic",
+            "crossover",
```
Luthaf, line 233: `crossover=True` is what rascaline does, `crossover=False` could be achieved once #134 is merged with a `selected_keys` parameter to the compute function

python/tests/misc.py
```diff
@@ -18,5 +20,91 @@ def test_cmake_files_exists(self):
...
+class TestConverter(unittest.TestCase):
+    """
+    Tests the hyperparameter conversions in `rascaline.utils`
+    """
+
+    def test_mode(self):
+        from rascaline.utils import convert_old_hyperparameter_names
+
+        with self.assertRaises(ValueError):
+            convert_old_hyperparameter_names({}, mode="BadMode")
```
Luthaf, line 32: please also check the error message when checking that exceptions are raised, otherwise it is really easy to get the tests to pass with a different error than the expected one beening raised.
```suggestion
        with self.assertRaises(ValueError) as cm:
            convert_old_hyperparameter_names({}, mode="BadMode")
        
        self.assertEqual(str(cm.exception), "some error message")
```

python/tests/misc.py
```diff
@@ -18,5 +20,91 @@ def test_cmake_files_exists(self):
...
+
+    def test_errant_params(self):
+        with self.assertRaises(ValueError):
+            convert_old_hyperparameter_names({"bad_param": 0}, mode="librascal")
+        with self.assertRaises(ValueError):
+            convert_old_hyperparameter_names({"bad_param": 0}, mode="dscribe")
+
+    def test_not_gto(self):
+        with warnings.catch_warnings(record=True) as w:
```
Luthaf, line 41: that works, but I've been using `self.assertWarns` everywhere else, so it might be better to use it here as well for consistency

Luthaf (commented):
This is very useful, thanks a lot! 

However I'm not sure if putting this function inside rascaline is the best thing to do from a marketing/future of the library point of view. I think this would be better as a semi-external tool (maybe a small website?) in which users can copy/paste hypers, forcing them to update there code. Right now I could see people keep the other packages syntax in their code, and use these function to dynamically translate to rascaline, while it would be better if they directly learn to work with rascaline hypers.

Alternatively, have this function print "your rascaline hypers are ..." as a string would work for me.

Luthaf:
linter is not happy, you can run `tox -e format` to reformat the code, and `tox -e lint` to run the flake8 checks locally

Luthaf:
I think this is worth having and tracked in #264, but in the mean I'll close this PR for inactivity. We can re-open something later!

PR #140, opened 2022-12-05
Provide pre-computed splines for the radial integrals

frostedoyster:


rascaline/src/calculators/radial_basis/mod.rs
```diff
@@ -21,6 +21,16 @@ pub enum RadialBasis {
         #[serde(default = "serde_default_spline_accuracy")]
         spline_accuracy: f64,
     },
+    Tabulated {
+        // Provide user-defined splines. 
+        // These consist of the positions of the spline points (as a serialized 1D array),
+        // values of the radial integrals at the spline point for each l and each n (as a 
+        // serialized flattened 3D array), and derivatives of the radial integrals at the spline 
+        // points for each l and each n (as another serialized flattened 3D array).
+        positions: String,
```
Luthaf, line 30: I think this would be better as a `Vec<f64>`, and have values/derivative as `Vec<Array2<f64>>` instead of adding more parsing down below. This could require a custom `deserialize_with` (https://serde.rs/field-attrs.html#serialize_with) and friends.

rascaline/src/math/splines.rs
```diff
@@ -35,18 +35,18 @@ pub struct SplineParameters {
...
+    pub value: Array<f64, D>,
     /// Derivative of the function to interpolate at the position
-    derivative: Array<f64, D>,
+    pub derivative: Array<f64, D>,
 }
 
 impl<D: ndarray::Dimension> HermitCubicSpline<D> {
     #[allow(clippy::float_cmp)]
-    fn new(
+    pub fn new(
```
Luthaf, line 49: This should not be public, since this is not something users of rascaline as a library should be able to use. It can be made public for other modules in rascaline with `pub(crate)` though.

A nicer solution might be to add a new public function that takes `Vec<f64>` for positions, and two `Vec<Array<f64, D>>` for values/gradient and move the validation (checking shapes, checking nothing is NaN/inf, …) inside this new function.

rascaline/src/calculators/radial_basis/mod.rs
Luthaf: Alternatively, this could also be 

```rust
pub enum RadialBasis {
    Gto {
        // ...  
    },
    Tabulated {
        points: Vec<SplinePoint>
    },
}


SplinePoint {
    positions: f64,
    #[serde(with=serialize_arrays)]
    values: Array2<f64>,
    #[serde(with=serialize_arrays)]
    derivatives: Array2<f64>,
}

mod serialize_arrays {
    todo!()
}
```

Luthaf:
The python one? You can create a new file and import it from `__init__.py` to make it accessible as `rascaline.create_spline`

python/examples/tabulated.py
```diff
@@ -0,0 +1,156 @@
...
+    return zerosj
+
+
+########################################################
+
+
+# Hyperparameters:
+lmax = 10
+nmax = 10
+a = 5.0  # Radius of the LE sphere - i.e, the cutoff radius
```
Luthaf, line 52: I would keep the notation consistent with everything else:

```suggestion
max_angular = 10
max_radial = 10
cutoff = 5.0  # This is also the radius of the LE sphere
```

python/examples/tabulated.py
```diff
@@ -0,0 +1,156 @@
...
+    return N_nl(n, el) * R * a ** (-1.5)
+
+
+# Quick normalization check:
+def random_LE_function(x):
+    return get_LE_function(1, 1, x) ** 2 * x**2
+
+
+normalization_check_integral, _ = sp.integrate.quadrature(random_LE_function, 0.0, a)
+print(f"Normalization check (needs to be close to 1): {normalization_check_integral}")
```
Luthaf, line 88:
```suggestion
normalization_check_integral, _ = sp.integrate.quadrature(lambda x: get_LE_function(1, 1, x) ** 2 * x**2, 0.0, a)
print(f"Normalization check (needs to be close to 1): {normalization_check_integral}")
```

python/examples/tabulated.py
```diff
@@ -0,0 +1,156 @@
...
+# The radial basis functions and their derivatives can be input into a spline
+# generator. This will output the positions of the spline points, the
+# values of the basis functions evaluated at the spline points, and the
+# corresponding derivatives.
+positions, values, derivatives = generate_splines(
+    get_LE_function, get_LE_function_derivative, nmax, lmax, a
+)
+
+# Feed the splines to the Rust calculator:
+# (IMPORTANT: "atomic_gaussian_width" and "cutoff_function" will be ignored)
```
Luthaf, line 113: `atomic_gaussian_width` I see why, but this is not the case for `cutoff_function` (or `radial_scaling`), since there are added after the radial integral evaluation

python/examples/tabulated.py
```diff
@@ -0,0 +1,156 @@
...
+    for r in r_array:
+        dummy_structures.append(ase.Atoms("CH", positions=[(0, 0, 0), (0, 0, r)]))
+    return dummy_structures
+
+
+r = np.linspace(0.1, 4.9, 20)
+structures = get_dummy_structures(r)
+spherical_expansion_coefficients = calculator.compute(structures)
+
+# Extract l = 0 features and check that the n = 2 predictions are the same:
```
Luthaf, line 146: why only the n=2?

python/rascaline/splines.py
```diff
@@ -0,0 +1,71 @@
...
+
+def generate_splines(
+    radial_basis,
+    radial_basis_derivatives,
+    max_radial,
+    max_angular,
+    cutoff_radius,
+    n_spline_points=101,
+):
+
```
Luthaf, line 13: This needs a docstring, to be added to the documentation, and unit tests

python/rascaline/splines.py
```diff
@@ -0,0 +1,71 @@
...
+    positions = []
+    for single_x in spline_x:
+        positions.append(single_x)
+
+    values = []
+    for single_f in spline_f:
+        values.append({})
+        values[-1]["v"] = 1
+        values[-1]["dim"] = [max_angular + 1, max_radial]
+        values[-1]["data"] = single_f.flatten().tolist()
```
Luthaf, line 62:
```suggestion
        values.append({
            # this is the data representation used by ndarray through serde
            "v": 1,
            "dim": [max_angular + 1, max_radial],
            "data" = single_f.flatten().tolist(),
        })
```

python/examples/tabulated.py
frostedoyster: Really? I thought I had tried changing them at some points and they seemed to have no effect. I will try again

python/examples/tabulated.py
frostedoyster: Just a random check. We can add more n values. It would be more difficult to do l>0 checks because one would need to divide by the spherical harmonics, which are not constant anymore in that case

python/rascaline/splines.py
```diff
@@ -0,0 +1,71 @@
...
+import numpy as np
+
+
+def generate_splines(
+    radial_basis,
+    radial_basis_derivatives,
+    max_radial,
+    max_angular,
+    cutoff_radius,
+    n_spline_points=101,
```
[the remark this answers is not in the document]
frostedoyster: How would that work? The functions can only be evaluated from Python, but the splining is done in the Rust part. I can't see any easy solutions

rascaline/src/calculators/radial_basis/mod.rs
```diff
@@ -21,8 +21,28 @@ pub enum RadialBasis {
...
+    Tabulated {
+        // Provide user-defined splines. 
+        // These consist of the positions of the spline points (as a serialized 1D array),
+        // values of the radial integrals at the spline point for each l and each n (as a 
+        // serialized flattened 3D array), and derivatives of the radial integrals at the spline 
+        // points for each l and each n (as another serialized flattened 3D array).
+        positions: Vec<f64>,
+        values: Vec<Array2<f64>>,
+        derivatives: Vec<Array2<f64>>,
+    },
```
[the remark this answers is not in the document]
frostedoyster: How difficult would implementing the json-to-SplinePoint conversion be?

python/rascaline/splines.py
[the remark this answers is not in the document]
Luthaf: We would have to re-implemenent spline evaluation in Python, it is a pretty easy algorithm

rascaline/src/calculators/radial_basis/mod.rs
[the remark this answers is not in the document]
Luthaf: Not sure what you mean? We can add a `#[derive(serde::Serialize, serde::Deserialize)]` on the `SplinePoint` struct to get json conversions for free.

python/examples/tabulated.py
frostedoyster: You're right. I was hallucinating: both radial scaling and the cutoff function are not ignored

rascaline/src/calculators/radial_basis/mod.rs
```diff
@@ -21,8 +21,28 @@ pub enum RadialBasis {
         #[serde(default = "serde_default_spline_accuracy")]
         spline_accuracy: f64,
     },
+    Tabulated {
```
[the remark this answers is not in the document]
frostedoyster: Yes, indeed. I've changed it now. However, we will still need to rethink all of this when we do the new hypers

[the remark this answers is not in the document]
Luthaf: Very true 😃

python/examples/tabulated.py
```diff
@@ -0,0 +1,160 @@
...
+    def function_to_integrate_to_get_normalization_factor(x):
+        return j_l(el, x) ** 2 * x**2
+
+    integral, _ = sp.integrate.quadrature(
+        function_to_integrate_to_get_normalization_factor, 0.0, z_nl[n, el]
+    )
+    return (1.0 / z_nl[n, el] ** 3 * integral) ** (-0.5)
+
+
+def get_LE_function(n, el, r):
```
Luthaf, line 75:
```suggestion
def lagrangian_eigenvalues_basis(n, el, r):
```

Luthaf (commented):
A good tool if you want to get improvement suggestions on your rust code is [clippy](https://github.com/rust-lang/rust-clippy/). After installation, you can run it with `cargo clippy`

python/rascaline/splines.py
```diff
@@ -0,0 +1,254 @@
...
+            value_evaluator_3D,
+            derivative_evaluator_3D,
+            requested_accuracy,
+        )
+        spline_x, spline_f, spline_df = dynamic_spliner.spline()
+
+    # Due to the inner workings of rascaline,
+    # values and derivatives need to be multiplied by 1/(4pi):
+    spline_f = spline_f / (4.0 * np.pi)
+    spline_df = spline_df / (4.0 * np.pi)
```
Luthaf, line 71: Ok, we need to resolve this before merging this PR. I'm fine with putting the 4pi factor either inside or outside of the radial integral definition, but we need to document what's expected as a radial integral. @kvhuguenin do you have a mathematical expression we can put in the documentation?

Luthaf (changes requested):

python/examples/tabulated.py
```diff
@@ -0,0 +1,158 @@
...
+
+def Jn_zeros(n, nt):
+    zerosj = zeros((n + 1, nt), dtype=np.float64)
+    zerosj[0] = arange(1, nt + 1) * pi
+    points = arange(1, nt + n + 1) * pi
+    racines = zeros(nt + n, dtype=np.float64)
+    for i in range(1, n + 1):
+        for j in range(nt + n - i):
+            foo = brentq(Jn, points[j], points[j + 1], (i,))
+            racines[j] = foo
```
Luthaf, line 40:
```suggestion
            racines[j] = brentq(Jn, points[j], points[j + 1], (i,))
```

is `racine` a word in english as well? This is the term for zero of a function in french =)

i.e. should this be `roots`?

Luthaf:
I implemented the schemars traits, so now the documentation should build!

python/examples/tabulated.py
frostedoyster: You're right. This is low-quality code from the scipy cookbook. It seems that although racine can be used as a synonym for root in all other contexts, that is not the case in math, so I'll change it

python/rascaline/splines.py
frostedoyster: Is it good now?

frostedoyster: @Luthaf

python/tests/splines.py
```diff
@@ -0,0 +1,55 @@
...
+                np.array(spline_points[0]["values"]["data"]), np.zeros(((9 + 1) * 12,))
+            )
+        )  # check values at r = 0.0
+        self.assertTrue(
+            np.allclose(
+                np.array(spline_points[0]["derivatives"]["data"]),
+                np.ones(((9 + 1) * 12,)) / (4.0 * np.pi),
+            )
+        )  # check derivatives at r = 0.0
+
```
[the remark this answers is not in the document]
frostedoyster: Should be done

python/rascaline/splines.py
Luthaf: yep!

Luthaf (approved):

PR #144, opened 2022-12-15
Spherical Bessel functions from scipy

frostedoyster:
This should be all the math we need for the LE basis


Luthaf:
Can you rebase this branch? The test error should be fixed by #145.

PR #145, opened 2022-12-16
Use tox major version as additional key for cache

PicoCentauri:
With tox version 4 released our caching in the Github action does not work anymore because we created them with tox version 3.

To fix this this PR adds the tox major version as an additinal key for the Github action caching.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -82,6 +82,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install tox
+          echo "tox_version=$(tox --version | awk '{split($0, s," "); print s[1]}')" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -82,6 +82,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install tox
+          echo "TOX_MAJOR=$(tox --version | awk '{split($0, s,"."); print s[1]}')" >> $GITHUB_ENV
```
Luthaf, line 85: I would do the full version instead of just the major part. having to recreate the cache a bit more often is fine for me, and should help preventing this kind of issues later.

PicoCentauri: fine for me.

PR #148, opened 2023-01-09
Add basic linting to prevent commiting debug code in rust

Luthaf:


Luthaf:
Thanks for the review!

PR #151, opened 2023-01-11
Change the order of samples in spherical exansion by pair

Luthaf:
This is a follow-up to #147. 

Previously, the samples would look like this

```
structure, pair_id, first_atom, second_atom
   0          1         1           0
   0          1         0           1
   0          2         1           2
   0          2         2           1
   1          1         1           0
   1          1         0           1
                ...
   0         -1         0           0
   0         -1         1           1
   0         -1         2           2
   1         -1         0           0
   1         -1         1           1
```

i.e; with the "self" pairs at the end of the samples, meaning there could be a jump in the "structure" sample. This changes it to group all structures, placing the "self" pairs first inside the samples for a given structure.

### DIFF
--- a/rascaline/src/calculators/neighbor_list.rs
+++ b/rascaline/src/calculators/neighbor_list.rs
@@ -294,49 +294,37 @@ impl FullNeighborList {
         return Ok(keys.finish());
     }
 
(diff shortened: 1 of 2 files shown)


ceriottm:
Works as advertised!

PR #153, opened 2023-01-27
Port bessel functions from scipy

Luthaf:
(opening a draft PR in case someone wants to help bring this to completion)

This PR adds most of the bessel functions to rascaline, porting them from scipy. These functions will be needed for a native implementation of the LE basis, and to implement numerical integration in the radial integral.

The main thing missing is a set of tests for the `jv`  and `yv` bessel functions


PicoCentauri:
This is very useful for the LE basis and the monomials!

Are we exporting these to Python as well? We need these functions for the tabulated splining if we don't want to add `scipy` as an additional dependency. Unfortunately, we would also need the derivatives as defined i.e. in [scipy.special.spherical_jn.](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.spherical_jn.html)

Luthaf:
Exporting these to Python was not really the plan since we are not in the business of providing special functions implementations (and actually most of this code would be better in a separate `special-fn` crate). I would rather add an optional dependency on scipy for this.

The idea for these was more to provide "creation of arbitrary radial basis splines" directly on the Rust side, or a pure Rust implementation of the LE basis.

PicoCentauri:
Okay I see. I am asking because I am doing a general `SplinedRadialBasis` class in Python which will include LE and monomials.

PR #154, opened 2023-02-02
Add an AtomicComposition calculator

PicoCentauri:
This PR adds a calculator to compute the the stochiometry of a system. To implement this I also add a `SamplesPerAtom` Labels build which simple creates a Label for each atom in the system, filtered by a type.

Also some minor formatting issues and typos are resolved within this PR.

----
:books: Documentation preview :books:: https://rascaline--154.org.readthedocs.build/en/154/


rascaline/src/calculators/composition.rs
```diff
@@ -0,0 +1,169 @@
...
+use crate::labels::{CenterSpeciesKeys, KeysBuilder};
+
+#[derive(Debug, Clone)]
+#[derive(serde::Deserialize, serde::Serialize, schemars::JsonSchema)]
+/// An composition calculator for obtaining the stoichiometric information.
+pub struct Composition {}
+
+impl CalculatorBase for Composition {
+    fn name(&self) -> String {
+        return "atom centered composition features".into();
```
frostedoyster, line 19:
```suggestion
        return "atom-centered composition features".into();
```

rascaline/src/calculators/composition.rs
```diff
@@ -0,0 +1,169 @@
...
+            array.fill(1.0);
+
+            if let Some(gradient) = block.gradient_mut("positions") {
+                let array = gradient.data.as_array_mut();
+                array.fill(0.0);
+            } else if let Some(gradient) = block.gradient_mut("cell") {
+                let array = gradient.data.as_array_mut();
+                array.fill(0.0);
+            }
+        }
```
frostedoyster, line 101: Are these necessary? Aren't all Rust arrays already initialized (I assume to zero)?

rascaline/src/calculators/composition.rs
```diff
@@ -0,0 +1,169 @@
...
+
+        return Ok(gradient_samples);
+    }
+
+    fn components(&self, keys: &Labels) -> Vec<Vec<Arc<Labels>>> {
+        return vec![Vec::new(); keys.count()];
+    }
+
+    fn properties_names(&self) -> Vec<&str> {
+        return vec!["count"];
```
frostedoyster, line 75: I think there might be an issue here. A single "count" property label does not make this kind of descriptor easily summable over different atoms in the same structure. To do that, in my opinion, we would instead need a "species" property label, which has dimension n_species (i.e., the properties dimension is n_species long) and which has value 1 for the species of the central atom and zero otherwise. This would make it summable across atoms in the same structure but having a different central atom species.

frostedoyster (changes requested):
Great work. I think we need a Python example, and especially a Python example where we use the equistore sum-over-structures utility function to generate composition features for a real system. Besides being very helpful to the user, I think this would bring to the surface the "count" vs "species" problem I mentioned

ceriottm:
Is this meant to streamline the "composition models"? Then effectively this is a structure property. It can be implemented either by having a property that is 1 for all centers, and have a sample index that indicates the central atom type (so that sum_over_samples will give the number of atoms of each type for a given structure) or by having a one-hot encoding on the property level as @frostedoyster suggests. I think the former is more data efficient and more in line with the ACDC concept.

PicoCentauri:
Tue current implementation is the first one you mentioned @ceriottm. 

I planned to introduce a 'structure_wise' parameter that will already perform the sum like 'sum_over_samples'

rascaline/src/calculators/composition.rs
PicoCentauri: Yes, you are right. We can leave them as they are!

PicoCentauri:
@frostedoyster  and @ceriottm. I added the `per_structure` parameter. Let me know if this is what you had in mind. If you agree I still have to fix the gradients and clean up the code.

I have the feeling that the two new label operators `SamplesPerAtom ` and `Structure`  are bit of an overkill but maybe @Luthaf  has an idea.

Luthaf:
> I have the feeling that the two new label operators SamplesPerAtom and Structure are bit of an overkill but maybe @Luthaf has an idea.

I did not look at the code, but this code can also live inline inside the calculator. There is no need to create a separate implementation if it is not going to be re-used by another calculator.

PicoCentauri:
> I did not look at the code, but this kind of code can also live inline inside the calculator. There is no need to create a separate implementation if it is not going to be re-used by another calculator.

Yes, maybe this is good idea. And if we need this in another, we can split this out again.

rascaline/src/calculators/atomic_composition.rs
```diff
@@ -0,0 +1,269 @@
...
+                        if species_center_key.i32() == species_center_sys {
+                            builder.add(&[system_i, system_i, center_i]);
+                        }
+                    }
+                }
+            } else {
+                for (sample_i, [structure_i, center_i]) in samples.iter_fixed_size().enumerate() {
+                    builder.add(&[sample_i, structure_i.usize(), center_i.usize()]);
+                }
+            }
```
PicoCentauri, line 105: I am bit puzzled here:

The code seems to work properly. But, the two logic branches give the same number of gradients with exactly the same sample names. Is this correct for the case of an atomic composition, that the gradients are the same regardless if I request a `per_atom` or `per_structure` descriptor?

If yes, we could remove the `self.per_structure==false` branch and simplify the code even further.

frostedoyster: Hmmm are you sure? Without experimenting with your code, I would expect the gradient samples to be structure, center, atom in the normal case and structure, atom in the per_structure case. Does it make sense to you? Can anybody confirm?

Luthaf: Anyway, both gradients samples should be empty here. The positions/cell gradients of the composition are zero everywhere ...

rascaline/src/calculator.rs
```diff
@@ -484,6 +485,7 @@ macro_rules! add_calculator {
 // [calculator-registration]
 static REGISTERED_CALCULATORS: Lazy<BTreeMap<&'static str, CalculatorCreator>> = Lazy::new(|| {
     let mut map = BTreeMap::new();
+    add_calculator!(map, "atomiccomposition", AtomicComposition);
```
Luthaf, line 488:
```suggestion
    add_calculator!(map, "atomic_composition", AtomicComposition);
```

python/rascaline/calculators.py
```diff
@@ -256,6 +256,23 @@ def compute(
...
+
+    For ``per_structure=True`` a sum for each structure is performed. The number of
+    atoms per structure is saved. The only sample left is names ``structure``.
+    """
+
+    def __init__(self, per_structure):
+        paramteres = {
+            "per_structure": per_structure,
+        }
+        super().__init__("atomiccomposition", paramteres)
```
Luthaf, line 273:
```suggestion
        super().__init__("atomic_composition", parameters)
```

python/rascaline/calculators.py
```diff
@@ -256,6 +256,23 @@ def compute(
...
+
+    For ``per_structure=False`` calculator has one property ``count`` that is
+    ``1`` for all centers, and has a sample index that indicates the central atom type.
+
+    For ``per_structure=True`` a sum for each structure is performed. The number of
+    atoms per structure is saved. The only sample left is names ``structure``.
+    """
+
+    def __init__(self, per_structure):
+        paramteres = {
```
Luthaf, line 270:
```suggestion
        parameters = {
```

rascaline/src/calculators/atomic_composition.rs
```diff
@@ -0,0 +1,269 @@
...
+
+                    for &s in all_species {
+                        if s == species {
+                            n_species += 1.;
+                        }
+                    }
+                    array[[system_i, 0]] = n_species;
+                }
+            } else {
+                array.fill(1.0);
```
Luthaf, line 152: this will fail is the user selects more samples than what this calculator would produce by default (i.e. trying to compute the composition of CH4 with 5 samples for center_species C (4 of which should be 0) and 5 samples for center_species H (1 of which should be 0)

rascaline/src/calculators/atomic_composition.rs
```diff
@@ -0,0 +1,269 @@
...
+                for (system_i, system) in systems.iter_mut().enumerate() {
+                    let all_species = system.species()?;
+                    let mut n_species = 0.;
+
+                    for &s in all_species {
+                        if s == species {
+                            n_species += 1.;
+                        }
+                    }
+                    array[[system_i, 0]] = n_species;
```
Luthaf, line 149: This will fail with samples/properties selection. The user could remove the "count" property (would be stupid, but meh), or have samples in a different order than the systems.

rascaline/src/labels/samples/long_range.rs
```diff
@@ -31,7 +31,7 @@ impl SamplesBuilder for LongRangeSamplesPerAtom {
         for (system_i, system) in systems.iter_mut().enumerate() {
             let species = system.species()?;
 
-            // we want to take all centers matching `species_center` iff
+            // we want to take all centers matching `species_center` if
```
Luthaf, line 34: `iff` means `if and only if`, if it is not clear we can expand to that!

PicoCentauri: Ahh no it is fine. I just saw the two ff and assuming it is was a typo.

rascaline/src/calculators/atomic_composition.rs
PicoCentauri: So we do a loop over properties like

```rust
for (property_i, property) in values.properties.iter().enumerate() {
}
```

and only fill if  `property_i  == 0`?

rascaline/src/calculators/atomic_composition.rs
PicoCentauri: But, shouldn't this be taken of by the creation of the samples?

If no we do a loop like

```rust
for (sample_i, [system, center]) in values.samples.iter_fixed_size().enumerate() {
    for (property_i, property) in values.properties.iter().enumerate() {
        array[[sample_i, property_i]] = 1.0;
    }
}
```

Luthaf: So this might need to be better documented, but the `samples` function returns the **default** set of samples. 

The actual ones for a given calculation might differ if the users passed something as `selected_samples` to `Calculator::compute`

Luthaf: and yes, that's the typical loop one has to do, see all the other calculators

rascaline/src/calculators/atomic_composition.rs
PicoCentauri: Se we do not initialize them at all?

rascaline/src/calculators/atomic_composition.rs
PicoCentauri: Okay, I changed the code. Looks still quite clean and simple.

rascaline/src/calculators/atomic_composition.rs
Luthaf: Yes we can return an empty set of samples

PicoCentauri: Hmm but if I do something like

```rust
    fn positions_gradient_samples(
        &self,
        keys: &Labels,
        samples: &[Arc<Labels>],
        systems: &mut [Box<dyn System>],
    ) -> Result<Vec<Arc<Labels>>, Error> {

        debug_assert_eq!(keys.count(), samples.len());
        let gradient_samples = Vec::new();
        return Ok(gradient_samples);
    }
```

The tests for finite difference of course don't work anymore. So, I am not sure what you mean.

rascaline/src/calculators/atomic_composition.rs
```diff
@@ -0,0 +1,277 @@
...
+        assert_eq!(descriptor.keys().names(), ["species_center"]);
+
+        for (key, mut block) in descriptor.iter_mut() {
+            let species_center = key[0].i32();
+
+            let values = block.values_mut();
+            let array = values.data.as_array_mut();
+
+            for property_i in 0..values.properties.size() {
+                if property_i == 0 {
```
Luthaf, line 141:
```suggestion
            for (property_i, &[count]) in values.properties.iter_fixed_size().enumerate() {
                if count == 0 {
```

rascaline/src/calculators/atomic_composition.rs
Luthaf: Sorry, needs to be a vector of empty labels (one for each key). Something like this should work

```rs
let gradient_samples = Arc::new(Labels::empty(["sample", "structure", "atom"]));
return Ok(vec![gradient_samples; keys.count()]);
```

PicoCentauri: Ahh I see, yes this should do the job.

PicoCentauri: However, this causes problems when running `sum_over_samples`. This leads to an `ValueError: cannot reshape array of size 0 into shape (0,newaxis)` because the gradients are empty. Also I am not sure what happening when applying `join`.

Luthaf: I would say any error like this is a bugs in `sum_over_samples`/`join`/... which I would rather fix there

frostedoyster (commented):
Good to go

PR #155, opened 2023-02-06
Use glibc==2.28, rust==1.59, update GH actions and use py3.10 in CI

PicoCentauri:
As written in the title this updates the CI scripts to ensure them to work properly. To ensure this we have to bump the versions of glibc and rust.


agoscinski (approved):
Equivalent changes as in https://github.com/lab-cosmo/equistore/pull/111

PR #158, opened 2023-02-16
RFC: Python function for general Power Spectrum

PicoCentauri:
This PR ports the `compute_power_spectrum` function from [equistore-examples](https://github.com/lab-cosmo/equistore-examples/blob/8cec11e9cae3c43646ddbcb1a660e2c02a051451/utils/models/soap.py#L10).

A general rust version of the code based on the `SoapPowerSpectrum` is the ultimate goal but I think this function is a valid start. There are still three issues to be resolved.

1. Explicit and implicit second parameter of the same descriptor leads to different results.
    In details the number of properties are different.
3. Order of the properties is not the same as in `SoapPowerSpectrum` calculator
4. Returned property names are different compared to the `SoapPowerSpectrum`. The latter uses a convention without
    underlines like `n1`, `n2`. The `compute_power_spectrum` uses underlines like `n_1` and `n_2`.
    We should resolve this inconsistency. I would go for the latter to be consistent with the other properties like  
    `species_neighbor_1` etc…


PicoCentauri:
I did some benchmarking and the rust version is between 6–10 times faster. I think the current timing is reasonable but shows us that the a more general rust implementation should be still thought of.

PicoCentauri:
Did we change something with in the `cmake` requirements?

```
  CMake Error at CMakeLists.txt:6 (cmake_minimum_required):
    CMake 3.16 or higher is required.  You are running version 3.10.2
```

Luthaf:
> Did we change something with in the cmake requirements?

yes, in https://github.com/lab-cosmo/equistore/pull/267, which is then required for the torch side of things

python/rascaline/power_spectrum.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def compute_power_spectrum(
+    spherical_expansion_1: TensorMap, spherical_expansion_2: TensorMap = None
+) -> TensorMap:
+    """Generate the rotationally invariant power spectrum from spherical expansions.
+
+    If two different sets of spherical expansions are provided, it is assumed that the
+    invariants are generated by taking quadratic combinations of those. Otherwise,
+    invariants are generated by combining coefficients of one spherical expansion.
```
Luthaf, line 22: This is not extremely clear to me. How about putting the corresponding equation here? 

Also, the condition is somehow wrong: it is more about `spherical_expansion_2` being None than the two of them being the same, i.e. `compute_power_spectrum(spx, spx)` will still use the second code path.

python/rascaline/power_spectrum.py
```diff
@@ -0,0 +1,201 @@
...
+
+import numpy as np
+from equistore import Labels, TensorBlock, TensorMap
+
+
+try:
+    from equistore import slice_block
+except ImportError as err:
+    raise ImportError(
+        "equistore-operations are required to use the power spectrum."
```
Luthaf, line 11:
```suggestion
        "the equistore-operations package is required to use the power spectrum"
```

python/rascaline/power_spectrum.py
```diff
@@ -0,0 +1,201 @@
...
+    ) from err
+
+
+def compute_power_spectrum(
```
Luthaf, line 15: Do we want this to be a single function, or do we want to follow the same `Calculator` interface as the rest of the code?

python/rascaline/power_spectrum.py
```diff
@@ -0,0 +1,201 @@
...
+
+    If two different sets of spherical expansions are provided, it is assumed that the
+    invariants are generated by taking quadratic combinations of those. Otherwise,
+    invariants are generated by combining coefficients of one spherical expansion.
+
+    .. note::
+        Currently only supports ``values`` and gradients with respect to ``positions``.
+
+    :param spherical_expansion_1: first spherical expansions
+    :param spherical_expansion_2: second spherical expansions
```
Luthaf, line 28:
```suggestion
    :param spherical_expansion_1: first spherical expansions coefficients
    :param spherical_expansion_2: second spherical expansions coefficients
```

python/rascaline/power_spectrum.py
```diff
@@ -0,0 +1,201 @@
...
+            raise ValueError(
+                "key names of spherical_expansion_1 are "
+                f"{spherical_expansion_1.keys.names} but have to be {key_names}"
+            )
+
+    blocks = []
+    keys = []
+
+    for (l1, sc1, sn1), block1 in spherical_expansion_1:
+        # Find that block indices that have tge same spherical_harmonics_l and
```
Luthaf, line 57:
```suggestion
        # Find that block indices that have the same spherical_harmonics_l and
```

python/rascaline/power_spectrum.py
Luthaf: Another point to mention here is that in the single spx case there will be only half of the blocks, since (ns1, ns2) and (ns2, ns1) are symmetric.

python/rascaline/power_spectrum.py
```diff
@@ -0,0 +1,201 @@
...
+                        properties1.tolist() + properties2.tolist()
+                        for properties1 in spx1.properties
+                        for properties2 in spx2.properties
+                    ],
+                    dtype=np.int32,
+                ),
+            )
+
+            # Compute the invariants by summation and store the results
+            # operation: ima, imb -> iab
```
Luthaf, line 114:
```suggestion
            # this is equivalent to an einsum with: ima, imb -> iab
```

python/rascaline/power_spectrum.py
```diff
@@ -0,0 +1,201 @@
...
+
+                if len(gradient1.samples) == 0 or len(gradient2.samples) == 0:
+                    continue
+
+                # The "sample" dimension in the power spectrum gradient samples do not
+                # necessarily matches the "sample" dimension in the spherical expansion
+                # gradient samples. We create new samples by creating a union between
+                # the two gradient samples.
+                gradients_samples_values = np.unique(
+                    np.concatenate([gradient1.samples, gradient2.samples])
```
Luthaf, line 143: I think this is wrong, and has the same issue we recently fixed in the Rust implementation

spx1 samples could contain atom 1 at position 0, while spx2 samples could contain atom 1 at position 2 (if the neighbor species are different in the single spx case, or just with different cutoffs if we give this function two separate spx)

This means the gradient sample (0, 0, 1) in block 1 and (2, 0, 1) in block 2 would actually refer to the same gradient, and this would fail to find it. Similarly, I could see a case where identical gradient samples refer to different atoms/neighbors.

docs/src/references/api/python/calculators.rst
```diff
@@ -35,6 +35,11 @@ Available Calculators
     :show-inheritance:
 
 
+.. autoclass:: rascaline.power_spectrum.PowerSpectrum
+    :members:
+    :show-inheritance:
+
+
```
Luthaf, line 42: I don't know if we want to document this in the same way as the other calculators, since it does not support the same set of options. Maybe in a separate place?

python/rascaline/power_spectrum.py
```diff
@@ -0,0 +1,245 @@
...
+                    if len(common_samples) == 0:
+                        continue
+
+                    spx_1 = slice_block(block_1, axis="samples", labels=common_samples)
+                    spx_2 = slice_block(block_2, axis="samples", labels=common_samples)
+                else:
+                    spx_1 = block_1
+                    spx_2 = block_2
+
+                # Maybe we have to distinguis between if
```
Luthaf, line 140: What does this mean?

python/rascaline/power_spectrum.py
```diff
@@ -0,0 +1,245 @@
...
+        )
+
+        gradient_values = np.zeros(
+            [gradients_samples.shape[0], 3, len(new_block.properties)]
+        )
+
+        # this is equivalent to an einsum with: ixma, imb -> ixab
+        gradient_1_values = factor * np.matmul(
+            gradient_1.values.swapaxes(2, 3),
+            block_2.values[gradient_1.samples["sample"], np.newaxis, :, :],
```
Luthaf, line 209: that does not quite work, if the two blocks have different central species they will have different value samples, and we can not index block_2 with the samples of 1

PicoCentauri:
I updated the docs with an example and the equations. Let me know if we need anything more.

python/rascaline/utils/power_spectrum.py
```diff
@@ -0,0 +1,320 @@
...
+            \nabla p_{nl} = \nabla \rho_{nlm}^\dagger \cdot \nu_{nlm} +
+                            \rho_{nlm}^\dagger \cdot \nabla \nu_{nlm}
+
+        .. note::
+            Currently only supports gradients with respect to ``positions``.
+
+        If ``calculator_2=None`` invariants are generated by combining the coefficients
+        of the spherical expansion of ``calculator_1``. The power spectrum can
+        be calculated only for :py:class:`rascaline.SphericalExpansion` or
+        :py:class:`rascaline.LodeSphericalExpansion`.
```
Luthaf, line 42:
```suggestion
        of the spherical expansion of ``calculator_1``. The spherical expansions given 
        as input can only be :py:class:`rascaline.SphericalExpansion` or
        :py:class:`rascaline.LodeSphericalExpansion`.
```

python/rascaline/utils/power_spectrum.py
```diff
@@ -0,0 +1,320 @@
...
+
+        :param calculator_1: first calculator
+        :param calculator_1: second calculator
+        :raises ValueError: If other calculators as
+            :py:class:`rascaline.SphericalExpansion` or
+            :py:class:`rascaline.LodeSphericalExpansion` is used.
```
Luthaf, line 48:
```suggestion
        :raises ValueError: If other calculators than
            :py:class:`rascaline.SphericalExpansion` or
            :py:class:`rascaline.LodeSphericalExpansion` are used.
```

python/rascaline/utils/power_spectrum.py
```diff
@@ -0,0 +1,320 @@
...
+
+        >>> sr_calculator = rascaline.SphericalExpansion(**sr_hypers)
+        >>> lr_calculator = rascaline.LodeSphericalExpansion(**lr_hypers)
+
+        Construct the power spectrum calculators and compute the spherical expansion
+
+        >>> calculator = rascaline.utils.PowerSpectrum(sr_calculator, lr_calculator)
+        >>> power_spectrum = calculator.compute(atoms)
+
+        Results are stored as for any other calculator
```
Luthaf, line 102:
```suggestion
        The resulting invariants are stored as :py:class:`equistore.TensorMap` as for any other calculator
```

python/rascaline/utils/power_spectrum.py
```diff
@@ -0,0 +1,320 @@
...
+        gradient_1_values = factor * np.matmul(
+            gradient_1.values.swapaxes(2, 3),
+            block_2.values[gradient_1.samples["sample"], np.newaxis, :, :],
+        )
+
+        # Find correct indices of the 1st gradient values and fill the
+        # values in
+        _, grad1_sample_idxs, _ = np.intersect1d(
+            gradients_samples, gradient_1.samples, return_indices=True
+        )
```
Luthaf, line 291: I'm surprised this works with the latest version of equistore from #194? `gradients.samples` is no longer a np.ndarray

python/tests/utils/power_spectrum.py
```diff
@@ -0,0 +1,134 @@
...
+    ps_rust = ps_rust.keys_to_samples(["species_center"])
+    ps_rust = ps_rust.keys_to_properties(["species_neighbor_1", "species_neighbor_2"])
+    kernel_rust = np.dot(ps_rust[0].values, ps_rust[0].values.T)
+    assert_allclose(kernel_python, kernel_rust)
+
+
+def test_power_spectrum_gradients() -> None:
+    """Test that gradients are correct using finite differences."""
+    # TODO
+    pass
```
Luthaf, line 124: Do you want to finish this before merging?

Luthaf: There is a Rust framework for finite differences here: https://github.com/Luthaf/rascaline/blob/9c2eb78af9a1564765b8381ac4acdfdaa079e21f/rascaline/src/calculators/tests_utils.rs#L251

python/tests/test_utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+    for atom_i in range(len(system)):
+        for spatial in range(3):
+            system_pos = copy(system)
+            system_pos._positions[atom_i][spatial] += displacement / 2
+            updated_pos = calculator.compute(system_pos)
+
+            system_neg = copy(system)
+            system_neg._positions[atom_i][spatial] -= displacement / 2
+            updated_neg = calculator.compute(system_neg)
```
PicoCentauri, line 43: I am bit puzzled here. Updating the values does not work for whatever reasons. Also, the style of the code, especially `system_pos._positions` does only work with the SystemForTests object and not for `ase` or other objects.

Any ideas how to fix and improve without touching a lot of files?

Luthaf: Have you tried a deepcopy? Also, you could make this function require an `ase.Atoms` object and rely on it to do modifications. Or even write a new small implementation of the system class just for these tests

PicoCentauri: Deepcopy fails because parts of the c-code which stems from the baseclass can not be pickled.

We could do ase and only test these with all-deps test case. Should be fine.

python/rascaline/profiling.py
```diff
@@ -19,6 +19,7 @@ class Profiler:
 
         with rascaline.Profiler() as profiler:
             # run some calculations
+            pass
```
Luthaf, line 22:
```suggestion
            ...
```

I think this is valid Python, and looks a bit better for examples!

python/tests/test_utils.py
```diff
@@ -0,0 +1,69 @@
...
+    The implementation is simular to ``rascaline/src/calculators/tests_utils.rs``.
+
+    :param calculator: calculator used to compute the representation
+    :param system: Atoms object
+    :param displacement: distance each atom will be displaced in each direction when
+        computing finite differences
+    :param max_relative: Maximal relative error. ``10 * displacement`` is a good
+        starting point
+    :param atol: Threshold below which all values are considered zero. This should be
+        very small (1e-16) to prevent false positives (if all values & gradients are
```
Luthaf, line 27: The default value does not respect this doc comment =) Should we update it to be something like 1e-12 or 1e-16?

python/tests/utils/power_spectrum.py
```diff
@@ -0,0 +1,147 @@
...
+        "Gto": {},
+    },
+    "cutoff_function": {
+        "ShiftedCosine": {"width": 0.5},
+    },
+}
+N_SPECIES = len(np.unique(SystemForTests().species()))
+
+
+def se_calculator():
```
Luthaf, line 32:
```suggestion
def soap_calculator():
```

python/tests/utils/power_spectrum.py
```diff
@@ -0,0 +1,147 @@
...
+def lode_calculator():
+    hypers = HYPERS.copy()
+    hypers.pop("cutoff_function")
+    hypers["potential_exponent"] = 1
+
+    return rascaline.LodeSphericalExpansion(**hypers)
+
+
+def se():
+    return se_calculator().compute(SystemForTests())
```
Luthaf, line 45:
```suggestion
def soap():
    return soap_calculator().compute(SystemForTests())
```

python/tests/utils/power_spectrum.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def ps():
```
Luthaf, line 48:
```suggestion
def power_spectrum():
```

Luthaf (commented):
Look good, I have a final set of naming suggestions, trying to make everything more explicit

python/rascaline/profiling.py
PicoCentauri: Yes this is better!

PR #159, opened 2023-02-17
Added more details on meta data and gradient calculations

PicoCentauri:
I had several discussions this week about how to position gradients to be organized. It was especially unclear why the position gradients have **more samples** than the features themselves. I tried to tackle this confusion by expanding the docstring of the `compute` method. If we agree on the docstring of the `compute` method I will update the docs of the other languages.

Additionally, I updated the tutorial with details on how the metadata (`samples`, `components`, and `properties`) for a calculator is looking. These details may fit better inside an equistore tutorial but we are lacking any introductory docs so far in equistore... It also doesn't hurt if we have something here. If the tutorial is too detailed we can also move this.

# TODO
- [] Update compute methods of C/C++ and rust


SanggyuChong:
Hello, I took a look at this and I believe these additions to the docstring and tutorial does help a lot and make a difference. Thanks, @PicoCentauri !

> Note: Position gradients of an atom are computed with respect to all other atoms. To recover the force one has to accumulate all pairs associated with atom.

Here, I wanted to ask if this is with respect to all other atoms or all other atoms within the cutoff? From the raw tensorblock size, I thought it was only with respect to neighboring atoms within the cutoff distance.

It would also be nice to include some graphical representations of these matters in the tutorial to come, more so for the thing where the keys can be interpreted as samples or properties. In my experience, the approach only became clear after I got a personal explanation of it from one of the developers. It was not so intuitive when I tried to figure it out on my own. Perhaps such a graphical representation is already somewhat being captured in the first figure of the "core concepts of rascaline" section, but I think this could be done better with more in-depth explanations.

PicoCentauri:
> Here, I wanted to ask if this is with respect to all other atoms or all other atoms within the cutoff? From the raw tensorblock size, I thought it was only with respect to neighboring atoms within the cutoff distance.

Yes, within the cutoff. I added this in the latest version.

> It would also be nice to include some graphical representations of these matters in the tutorial to come, more so for the thing where the keys can be interpreted as samples or properties. In my experience, the approach only became clear after I got a personal explanation of it from one of the developers. It was not so intuitive when I tried to figure it out on my own. Perhaps such a graphical representation is already somewhat being captured in the first figure of the "core concepts of rascaline" section, but I think this could be done better with more in-depth explanations.

This would be more fitting for an equistore tutorial. But, yes we definetly need something like this.

docs/src/references/calculators/index.rst
```diff
@@ -17,6 +17,9 @@ with ``Calculator::new`` in Rust, ``rascal_calculator`` in C or
 given as a JSON formatted string. The possible fields in the JSON are documented
 as a `JSON schema`_, and rendered in the pages below.
 
+After the initialization of each calculator the computation of the
+representations is performed using the ``compute()`` method.
```
Luthaf, line 21: Should this be a link to the Python function?

python/examples/first-calculation.py
```diff
@@ -102,108 +102,213 @@
...
 # represented using their atomic number, e.g. 1 for hydrogen, 6 for carbon, etc.
-# Our descriptor contains 112 blocks covering all combinations of the angular
-# channels and the elements.
+# In sum descriptor contains 112 blocks covering all combinations of the angular
+# channels of the central and neighbor atom species in our dataset.
 #
-# Let us take a look at the first block in detail. The 0th block contains the
-# descriptor for the 0th angular channel between the hydrogen-hydrogen pairs.
+# Let us take a look at the second block in detail. The 1th block contains the
+# descriptor for the 1th angular channel between the hydrogen-hydrogen pairs.
```
Luthaf, line 138:
```suggestion
# Let us take a look at the second block (at index 1) in detail. This block contains the
# descriptor for the :math:`l=1` angular channel for hydrogen-hydrogen pairs.
```

python/examples/first-calculation.py
```diff
@@ -102,108 +102,213 @@
...
+print(descriptor0.keys[1])
+
+
+# %%
+# The descriptor values
+# ---------------------
+#
+# The values of the representation are stored as an array. Each entry in this
+# array also has associated unique metadata as each block.
+# For the SOAP calculator used in this tutorial the values have three dimensions
```
Luthaf, line 150:
```suggestion
# For the spherical expansion calculator used in this tutorial the values have three dimensions
```

python/examples/first-calculation.py
```diff
@@ -102,108 +102,213 @@
...
-# to positions, which is useful for constructing an ML potential and running
-# simulations. ``gradients``
-# of the representation with respect to atomic positions can be calculated by
-# setting the ``gradients`` parameter of the compute method to
-# ``["positions"]``.
+# Additionally, rascaline is also able to calculate gradients on top of the values.
+# Gradients are useful for constructing an ML potential and running simulations.
+# For example ``gradients`` of the representation with respect to atomic positions
+# can be calculated by setting the ``gradients`` parameter of the
+# :meth:`rascaline.calculators.CalculatorBase.compute()` method to ``["positions"]``.
```
Luthaf, line 250:
```suggestion
# :py:func:`rascaline.calculators.CalculatorBase.compute()` method to ``["positions"]``.
```

python/examples/first-calculation.py
```diff
@@ -102,108 +102,213 @@
...
+gradient_position = block_gradients.gradient("positions")
+
+print(gradient_position.data.shape)
+
+# %%
+#
+# The calculated descriptor contains the values and in each block the
+# associated position gradients as an :class:`equistore.block.Gradient` instance.
+# The actual values are stored in the ``data`` attribute. Similar to the features
+# the gradient data also has associated mete data. But, compared to the values were
```
Luthaf, line 264:
```suggestion
# the gradient data also has associated metadata. But, compared to the values were
```

python/examples/first-calculation.py
```diff
@@ -102,108 +102,213 @@
...
+# we found three dimensions, and gradients have four.
+# Again the first is called `samples` and the `properties`.
+# The dimensions between the sample and property dimensions are denoted by
+# `components`.
+#
+# Looking at the shape in more detail we find that we
+# have 52 samples, which is much more compared to features where we only
+# have eight samples. This arises from the fact that we calculate the position
+# gradient for each pair in the structure. For our selected block these are
+# all hydrogen-hydrogen pairs. Naively one would come up with `8*8=64`
```
Luthaf, line 274:
```suggestion
# all hydrogen-hydrogen pairs. Naively one would come up with ``8 * 8 = 64``
```

python/examples/first-calculation.py
```diff
@@ -102,108 +102,213 @@
...
+# Note that we have a tuple of three with the names
+
+print(gradient_position.samples.names)
+
+# %%
+#
+# In the above output of the Labels instance for example the `(2, 0, 17)` entry is
+# missing indicating that this pair is outside of the cutoff.
+#
+# Now looking at the :attr:`equistore.block.Gradient.components` we find
```
Luthaf, line 291:
```suggestion
# Now looking at the :attr:`equistore.block.Gradient.components`
```

PicoCentauri:
I copied the docstring over to C/C++/rust. You can do you formatting magic @Luthaf :-).

Luthaf (approved):

PR #170, opened 2023-03-17
Do not include equistore in python wheels

Luthaf:
We are always using libequistore.so from the corresponding Python package, and I just learned about the `--exclude` flag to `auditwheel repair`, which is intended for this exact use case.

----
:books: Documentation preview :books:: https://rascaline--170.org.readthedocs.build/en/170/


ceriottm:
What are the implications? End of "usage depends on order of installation" nightmare?

Luthaf:
unfortunately no, we will have some order of install issues until we can make releases with stability guarantees.

libequistore.so was included in the wheel only to satisfy auditwheel, and never used by the code. This will mainly reduce the size of the downloaded wheels (and remove a potential footgun when someone tries to use rascaline without having equistore installed)

PR #172, opened 2023-04-04
Improve usage of C library from installed Python package

Luthaf:
This feature is used in rascaline-torch. This PR does three things:

- move `find_package(equistore)` to a later stage, otherwise it overwrites the `PACKAGE_PREFIX_DIR` variable
- do not include equistore cmake files in rascaline wheel
- add checks that the installation contains all the required files

----
:books: Documentation preview :books:: https://rascaline--172.org.readthedocs.build/en/172/


PicoCentauri (approved):
👍

PR #179, opened 2023-04-26
Update to the latest equistore

Luthaf:
- gradients are now full TensorBlocks
- rascaline only depends on equistore-core, not the full package

----
:books: Documentation preview :books:: https://rascaline--179.org.readthedocs.build/en/179/

### DIFF
--- a/rascaline/Cargo.toml
+++ b/rascaline/Cargo.toml
@@ -36,7 +36,7 @@ name = "soap-power-spectrum"
 harness = false
 
 [dependencies]
-equistore = {git = "https://github.com/lab-cosmo/equistore", rev = "6c5a833", features = ["rayon"]}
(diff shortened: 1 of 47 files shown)


rascaline/Cargo.toml
```diff
@@ -36,7 +36,7 @@ name = "soap-power-spectrum"
 harness = false
 
 [dependencies]
-equistore = {git = "https://github.com/lab-cosmo/equistore", rev = "6c5a833", features = ["rayon"]}
+equistore = {git = "https://github.com/lab-cosmo/equistore", rev = "e5b9dc365369ba2584ea01e9d6a4d648008aaab8", features = ["rayon"]}
```
PicoCentauri, line 39: I think the short hash would be enough but we will change it soon. lets keep it like this.

PicoCentauri (approved):
Cool, looks solid!

PR #199, opened 2023-07-17
Multiple improvements to prepare for rascaline-torch moving day

Luthaf:
This PR contains a handful of small changes/cleanup unrelated to rascaline-torch but that should happen before the code moves to it's new home.

----
:books: Documentation preview :books:: https://rascaline--199.org.readthedocs.build/en/199/

PR #200, opened 2023-07-20
Move and cleanup the code for rascaline-torch

Luthaf:
This imports all of https://github.com/Luthaf/rascaline-torch in this repo. The main change is that the code now uses equistore_torch for the return type, allowing full TorchScript integration.

This is built on top of #199 which should be merged first.

Still TODO:
- [ ] sample/properties/keys selection
- [x] add torch to wheels building tests
- [ ] gather code coverage for torch
- [x] test using rascaline-torch from CMake
- [x] incorporate #205 in this PR
- [x] enable double backward

Ping @bananenpampe since you are using the old code quite a bit, could you give this branch a try? It should work in mostly the same way

----
:books: Documentation preview :books:: https://rascaline--200.org.readthedocs.build/en/200/


tox.ini
```diff
@@ -73,21 +79,45 @@ commands =
...
+
+    cmake
+    wheel
+
+changedir = python/rascaline-torch
+commands =
+    # install rascaline-torch
+    pip install . {[testenv]build-single-wheel} --force-reinstall
+    # run the unit tests
+    pytest --import-mode=append --assert=plain {posargs}
```
PicoCentauri, line 106: @PicoCentauri check this if this can be removed.

Luthaf:
Ok, I think this is ready for review @PicoCentauri, @agoscinski @DavideTisi if you have some time to look into the code! 

In particular I'd like to check if everything is comprehensible and if you can follow what's going on.
Some tutorials on the torch side that might be helpful to understand this code: https://pytorch.org/tutorials/intermediate/custom_function_double_backward_tutorial.html, https://pytorch.org/tutorials/advanced/torch_script_custom_classes.html

rascaline-torch/src/calculator.cpp
```diff
@@ -0,0 +1,289 @@
+#include "rascaline/torch/calculator.hpp"
+#include "equistore/torch/tensor.hpp"
+#include "rascaline/torch/autograd.hpp"
+#include <c10/util/Exception.h>
+
+using namespace equistore_torch;
+using namespace rascaline_torch;
+
+// move a block created by rascaline to torch
+static equistore::TensorBlock block_to_torch(
```
PicoCentauri, line 10: Shouldn't this be in equistore?

tox.ini
PicoCentauri: I played around but didn't found anything. Lets keep this and maybe pytets will solve this in a future version.

rascaline-torch/src/calculator.cpp
Luthaf: We could move this to equistore-torch, but right now is a bit specific to rascaline, assuming f64 and contiguous data that can be kept alive just by keeping the TensorMap around. Once we have something like https://github.com/lab-cosmo/equistore/issues/56, we could support more use cases and move this code over there

python/rascaline-torch/tests/autograd.py
```diff
@@ -0,0 +1,241 @@
+import warnings
+
+import ase
+import pytest
+import torch
+
+import rascaline.torch
+from rascaline.torch import SoapPowerSpectrum, SphericalExpansion, System
```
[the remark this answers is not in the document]
Luthaf: You mean a basic system with minimal functionality? We already have [SystemBase](https://luthaf.fr/rascaline/latest/references/api/python/systems.html#rascaline.SystemBase) in rascaline for custom system, which can be used like rascaline-torch system is used.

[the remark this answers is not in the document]
PicoCentauri: Yes, but SystemBase is very rigid. The species, positions and cell is fixed. Here you can give this as arguments, which would make the import of `ase` in some more cases superflous.

rascaline-torch/src/calculator.cpp
PicoCentauri: Okay fine for me.

python/rascaline-torch/tests/autograd.py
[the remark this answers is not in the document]
Luthaf: > The species, positions and cell is fixed.

No? SystemBase is a base class, if you want to reproduce the functionality of the `rascaline.torch.System` you can do something like 

```py
class System(SystemBase):
    def __init__(self, species, positions, cell):
        self.species = species
        self.positions = positions
        self.cell = cell
        # TODO: size validation for all data

    def size(self):
        return len(self.species)

    def species(self):
        return self.species

    def positions(self):
        return self.positions

    def cell(self):
        return self.cell

    def compute_neighbors(self, cutoff):
        raise NotImplementedError("this can not be used with use_native_systems=True")

    def pairs(self):
        raise NotImplementedError("this can not be used with use_native_systems=True")

    def pairs_containing(self, center):
        raise NotImplementedError("this can not be used with use_native_systems=True")
```

We could have this code as an example in SystemBase, but it feels to me it will not be that useful, since most people will use the ASE or chemfiles systems implementation.

We can and should use things like this for tests, we already do it in some places: https://github.com/Luthaf/rascaline/blob/9a049175be29a679ad954ce83b577c3a5b2cdc1e/python/rascaline/tests/test_systems.py#L4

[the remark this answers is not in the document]
Luthaf: Added documentation in #217

PicoCentauri (approved):
Big leap forward! Very nice.

PR #202, opened 2023-07-21
Some consistency improvements to splining

PicoCentauri:
- Renaming `value` -> `values` and `derivative` -> `derivatives` in `spline.rs` to be consisted with the tabulated splining code.
 - Test that exporting the spline points from `SoapRadialIntegralSpline::with_accuracy` and loading them into `SoapRadialIntegralSpline::from_tabulated` works directly. This test does not only checks that the two `struct`s are consistent but also that to methods create the same radial integrals.
 - typehints in `spline.py` 

One thing that I found maybe improvable in the future is that we seem to use `gradients` and sometimes `derivatives` for the same thing. I try tried to improve this by using `derivatives` for everything that is related to the splines and `gradients` appears for the radial integral. But, maybe we can call everything `gradients`?

----
:books: Documentation preview :books:: https://rascaline--202.org.readthedocs.build/en/202/

### DIFF
--- a/python/rascaline/rascaline/splines.py
+++ b/python/rascaline/rascaline/splines.py
@@ -1,28 +1,29 @@
+from typing import Callable, Optional
+
 import numpy as np
 
 
 def generate_splines(
-    radial_basis,
-    radial_basis_derivatives,
-    max_radial,
-    max_angular,
-    cutoff_radius,
-    n_spline_points=None,
-    requested_accuracy=1e-8,
+    radial_basis: Callable[[int, int, np.ndarray], np.ndarray],
+    radial_basis_derivatives: Callable[[int, int, np.ndarray], np.ndarray],
+    max_radial: int,
+    max_angular: int,
+    cutoff_radius: float,
+    n_spline_points: Optional[int] = None,
+    requested_accuracy: float = 1e-8,
 ):
     """Spline generator for tabulated radial integrals.
 
     Besides some self-explanatory parameters, this function takes as inputs two
-    functions, namely radial_basis and radial_basis_derivatives. These must be
-    able to calculate the radial basis functions by taking n, l, and r as their
-    inputs, where n and l are integers and r is a numpy 1-D array that contains
-    the spline points at which the radial basis function (or its derivative)
-    needs to be evaluated. These functions should return a numpy 1-D array
-    containing the values of the radial basis function (or its derivative)
-    corresponding to the specified n and l, and evaluated at all points in the
-    r array. If specified, n_spline_points determines how many spline points
-    will be used for each splined radial basis function. Alternatively, the user
-    can specify a requested accuracy. Spline points will be added until either
+    functions, namely radial_basis and radial_basis_derivatives. These must be able to
+    calculate the radial basis functions by taking n, l, and r as their inputs, where n
+    and l are integers and r is a numpy 1-D array that contains the spline points at
+    which the radial basis function (or its derivative) needs to be evaluated. These
+    functions should return a numpy 1-D array containing the values of the radial basis
+    function (or its derivative) corresponding to the specified n and l, and evaluated
+    at all points in the r array. If specified, n_spline_points determines how many
+    spline points will be used for each splined radial basis function. Alternatively,
+    the user can specify a requested accuracy. Spline points will be added until either
     the relative error or the absolute error fall below the requested accuracy on
     average across all radial basis functions.
     """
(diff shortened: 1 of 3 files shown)


python/rascaline/rascaline/splines.py
```diff
@@ -1,28 +1,29 @@
...
 
 def generate_splines(
-    radial_basis,
-    radial_basis_derivatives,
-    max_radial,
-    max_angular,
-    cutoff_radius,
-    n_spline_points=None,
-    requested_accuracy=1e-8,
+    radial_basis: Callable,
```
Luthaf, line 7: I think you can even define arguments and return values with `Callable`

python/rascaline/rascaline/splines.py
```diff
@@ -1,28 +1,29 @@
...
+    radial_basis_derivatives: Callable,
+    max_radial: int,
+    max_angular: int,
+    cutoff_radius: float,
+    n_spline_points: Optional[int] = None,
+    requested_accuracy: Optional[float] = 1e-8,
```
Luthaf, line 13:
```suggestion
    requested_accuracy: float = 1e-8,
```

`Optional[T]` means `Union[None, T]`

python/rascaline/rascaline/splines.py
```diff
@@ -87,26 +88,25 @@ def derivative_evaluator_3D(positions):
...
-        start,
-        stop,
-        values_fn,
-        derivatives_fn,
-        requested_accuracy,
+        start: float,
+        stop: float,
+        values_fn: Callable,
+        derivatives_fn: Callable,
+        requested_accuracy: Optional[float] = 1e-8,
```
Luthaf, line 95:
```suggestion
        requested_accuracy: float = 1e-8,
```

rascaline/src/math/splines.rs
```diff
@@ -9,21 +9,21 @@ const MAX_SPLINE_SIZE: usize = 10_000;
...
+#[derive(Debug, Clone, serde::Serialize)]
 pub struct HermitCubicSpline<D: ndarray::Dimension> {
     parameters: SplineParameters,
     points: Vec<HermitSplinePoint<D>>,
 }
 
 
 /// Parameters controlling a `HermitCubicSpline`
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, serde::Serialize)]
```
Luthaf, line 26:
```suggestion
#[derive(Debug, Clone)]
```

rascaline/src/math/splines.rs
```diff
@@ -34,14 +34,14 @@ pub struct SplineParameters {
 }
 
 /// A single control point/knot in the Hermit cubic spline
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, serde::Serialize)]
```
Luthaf, line 37:
```suggestion
#[derive(Debug, Clone)]
```

rascaline/src/calculators/soap/radial_integral/spline.rs
```diff
@@ -161,4 +162,55 @@ mod tests {
...
+
+        assert_relative_eq!(
+            values_accuracy, values_tabulated,
+            epsilon=1e-15, max_relative=1e-16
+        );
+
+        assert_relative_eq!(
+            gradients_accuracy, gradients_tabulated,
+            epsilon=1e-15, max_relative=1e-16
+        );
```
Luthaf, line 213: If we want to check these are exactly the same, we could use `assert_eq!` here instead of the relative_eq

rascaline/src/math/splines.rs
```diff
@@ -9,21 +9,21 @@ const MAX_SPLINE_SIZE: usize = 10_000;
...
-/// This kind of spline uses information of the value of a function on control
-/// points as well as the gradient of the function at these points. This
+/// This kind of spline uses information of the values of a function on control
+/// points as well as the derivatives of the function at these points. This
 /// implementation takes a single scalar as input, and output array values, i.e.
 /// we can spline functions of the form `R -> R^n`.
 ///
 /// [splines-wiki]: https://en.wikipedia.org/wiki/Cubic_Hermite_spline
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, serde::Serialize)]
```
Luthaf, line 18: I would not put `serde::Serialize` here, since it would force us to increase the version number if we ever change the names/types of the fields. For the test adding a `pub(crate) fn points(&self) -> &[HermitSplinePoint<D]` should be enough to then extract positions/values/derivatives and serialize the whole thing.

rascaline/src/calculators/soap/radial_integral/spline.rs
PicoCentauri: I tried this before but, the values are not exactly the same...

rascaline/src/math/splines.rs
PicoCentauri: You want to add this function here? 

Isn't it then just enough to change the `points` below to `pub(crate) points`?

PicoCentauri:
I removed the Serialize from the splining. Let me know if this is what you had in mind.

rascaline/src/calculators/soap/radial_integral/spline.rs
Luthaf: ok, fair enough

rascaline/src/math/splines.rs
Luthaf: yeah, that works as well ^^

Luthaf:
Looks fine, tests should pass after a rebase!

Luthaf (approved):
Feel free to merge once CI is happy

PR #204, opened 2023-07-21
Enable splines for LODE

PicoCentauri:
This is a WIP work and currently works but doesn't support the `central_atom_contribution`

### DIFF
--- a/rascaline/src/calculators/lode/radial_integral/spline.rs
+++ b/rascaline/src/calculators/lode/radial_integral/spline.rs
@@ -1,7 +1,8 @@
 use ndarray::{Array1, Array2, ArrayViewMut2};
 
 use super::LodeRadialIntegral;
-use crate::math::{HermitCubicSpline, SplineParameters};
+use crate::math::{HermitCubicSpline, SplineParameters, HermitSplinePoint};
+use crate::calculators::radial_basis::SplinePoint;
 use crate::Error;
 
 /// `LodeRadialIntegralSpline` allows to evaluate another radial integral
@@ -63,6 +64,42 @@ impl LodeRadialIntegralSpline {
             center_contribution: radial_integral.compute_center_contribution()
         });
     }
+
+    /// Create a new `LodeRadialIntegralSpline` with user-defined spline points.
+    pub fn from_tabulated(
+        parameters: LodeRadialIntegralSplineParameters,
+        spline_points: Vec<SplinePoint>,
+        center_contribution: Vec<f64>,
+    ) -> Result<LodeRadialIntegralSpline, Error> {
+
+        let spline_parameters = SplineParameters {
+            start: 0.0,
+            stop: parameters.cutoff,
+            shape: vec![parameters.max_angular + 1, parameters.max_radial],
+        };
+
+        let mut new_spline_points = Vec::new();
+        for spline_point in spline_points {
+            new_spline_points.push(
+                HermitSplinePoint{
+                    position: spline_point.position,
+                    values: spline_point.values.0.clone(),
+                    derivatives: spline_point.derivatives.0.clone(),
+                }
+            );
+        }
+
+        if center_contribution.len() != parameters.max_radial {
+            return Err(Error::InvalidParameter(format!(
+                "wrong length of center_contribution, expected {} elements but got {}",
+                parameters.max_radial, center_contribution.len()
+            )))
+        }
+
+        let spline = HermitCubicSpline::new(spline_parameters, new_spline_points);
+        return Ok(LodeRadialIntegralSpline{
+            spline: spline, center_contribution: Array1::from_vec(center_contribution)});
+    }
 }
 
 impl LodeRadialIntegral for LodeRadialIntegralSpline {
@@ -79,6 +116,7 @@ impl LodeRadialIntegral for LodeRadialIntegralSpline {
 #[cfg(test)]
 mod tests {
     use approx::assert_relative_eq;
+    use ndarray::Array;
 
     use super::*;
     use super::super::{LodeRadialIntegralGto, LodeRadialIntegralGtoParameters};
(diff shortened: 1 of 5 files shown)


rascaline/src/calculators/lode/radial_integral/spline.rs
```diff
@@ -63,6 +64,33 @@ impl LodeRadialIntegralSpline {
...
+                    position: spline_point.position,
+                    values: spline_point.values.0.clone(),
+                    derivatives: spline_point.derivatives.0.clone(),
+                }
+            );
+        }
+
+        let spline = HermitCubicSpline::new(spline_parameters, new_spline_points);
+        return Ok(LodeRadialIntegralSpline{
+            spline: spline, center_contribution: Array1::zeros(parameters.max_radial)});
```
PicoCentauri, line 92: `center_contribution` is currently set to zero here, we will also have to provide these values with the splines something like:

```python
radial_basis={"TabulatedRadialIntegral": {"points": spline_points, "center_contribution": xxx}},
```

rascaline/src/calculators/radial_basis/mod.rs
```diff
@@ -28,7 +28,12 @@ pub enum RadialBasis {
     /// The easiest way to create a set of spline points is the
     /// `rascaline.generate_splines` Python function.
     TabulatedRadialIntegral {
+        /// spline points
         points: Vec<SplinePoint>,
+        /// contribution of the central atom. This parameter is only required for LODE
+        /// calculations where the length of `center_contribution` must be the same as
+        /// `max_radial`.
+        center_contribution: Option<Vec<f64>>,
```
PicoCentauri, line 36: I made the `center_contribution` an optional argument. Let me know if this is okay of if we should create a new enum entry like `TabulatedRadialIntegralCenter`

rascaline/src/calculators/lode/radial_integral/spline.rs
```diff
@@ -63,6 +64,40 @@ impl LodeRadialIntegralSpline {
...
+            new_spline_points.push(
+                HermitSplinePoint{
+                    position: spline_point.position,
+                    values: spline_point.values.0.clone(),
+                    derivatives: spline_point.derivatives.0.clone(),
+                }
+            );
+        }
+
+        assert_eq!(
```
Luthaf, line 91: I think this should be a `InvalidParameterError`, `assert_eq` will `panic`, which creates an `internal error (this is a bug)` error message

rascaline/src/calculators/lode/radial_integral/spline.rs
```diff
@@ -63,6 +64,40 @@ impl LodeRadialIntegralSpline {
             center_contribution: radial_integral.compute_center_contribution()
         });
     }
+
+    pub fn from_tabulated(
```
Luthaf, line 68: Please add a short doc comment here

rascaline/src/calculators/radial_basis/mod.rs
Luthaf: That's fine for me, but we need to document a bit better what "contribution of the central atom" means (i.e. what's the equation that should be computed)

rascaline/src/calculators/lode/radial_integral/spline.rs
PicoCentauri: Done!

PicoCentauri:
I added the equation and applied some extra math typestting.

rascaline/src/calculators/radial_basis/mod.rs
```diff
@@ -10,8 +10,8 @@ pub use self::tabulated::SplinePoint;
 pub enum RadialBasis {
     /// Use a radial basis similar to Gaussian-Type Orbitals.
     ///
-    /// The basis is defined as `R_n(r) ∝ r^n e^{- r^2 / (2 σ_n^2)}`, where `σ_n
-    /// = cutoff * \sqrt{n} / n_max`
+    /// The basis is defined as $R_n(r) \propto r^n \exp\left[ -r^2 / (2 σ_n^2) \right]$,
```
Luthaf, line 13: This is not rendered as math, so I think the previous code is better here: https://rascaline--204.org.readthedocs.build/en/204/references/calculators/spherical-expansion.html#json-schema-radialbasis

rascaline/src/calculators/radial_basis/mod.rs
```diff
@@ -28,7 +28,22 @@ pub enum RadialBasis {
...
+        ///        \sqrt{4 \pi} \int \mathrm{d}r r^2 R_n(r) g(r)$$
+        ///
+        /// where $g(r)$ is a radially symmetric density function, $R_n(r)$ the radial
+        /// basis function and $n$ the current radial channel. Note that the integration
+        /// range was deliberately left ambiguous since it depends on the radial basis,
+        /// i.e. for the GTO basis, $r \in \mathbb{R}^+$ is used, while
+        /// $r \in [0, r_\mathrm{cut}]$ for the monomial basis.
+        ///
+        /// This parameter is only required for LODE calculations where the length of
+        /// the provided `center_contribution` must be the same as `max_radial`.
```
Luthaf, line 45: Unfortunately, we don't render inner members documentation in the doc, could you move this above in the `TabulatedRadialIntegral` doc? Cf the link in my other comment for a rendered version.

rascaline/src/calculators/radial_basis/mod.rs
PicoCentauri: i only checked the rust docs and there it looks fine. Sure I can adjust this.

Luthaf: yeah, this is used in two places. We could (and probably should) improve the calculator reference to render math, but the code extracting the doc is a bit crude.

PicoCentauri: Yeah its fine. Now I know :-)

rascaline/src/calculators/radial_basis/mod.rs
```diff
@@ -27,8 +27,17 @@ pub enum RadialBasis {
...
     /// The easiest way to create a set of spline points is the
     /// `rascaline.generate_splines` Python function.
+    ///
+    /// For LODE calculations also the contribution of the central atom have to be
+    /// provided. The `center_contribution` is defined as `c_n =
+    /// \sqrt{4π} \int dr r^2 R_n(r) g(r)` where `g(r)` is a radially symmetric density
+    /// function, `R_n(r)` the radial basis function and `n` the current radial channel.
+    /// Note that the integration range was deliberately left ambiguous since it depends
+    /// on the radial basis, i.e. for the GTO basis, `r \in R^+` is used, while `r \in
+    /// [0, cutoff]` for the monomial basis.
```
Luthaf, line 37: I've seen that a couple of times, but I'm not sure I agree: if you want a smaller integration range, you can just define your basis to be zero outside the cutoff. And then the integration range is always `0, \infty`. I'm fine with leaving this here for now though, we can always revisit this when we update rascaline's hypers

Luthaf (approved):

PR #207, opened 2023-07-25
Expose key selections to the C++ interface

### LINKED ISSUE #142: Add support for selected_keys in the C++ API
#134 added the option to select keys in the output, and made the option available in Rust/C/Python. They now need to be added to the C++ API as well!

Luthaf:
Fix #142, will be used in #200

----
:books: Documentation preview :books:: https://rascaline--207.org.readthedocs.build/en/207/

### DIFF
--- a/rascaline-c-api/CMakeLists.txt
+++ b/rascaline-c-api/CMakeLists.txt
@@ -299,7 +299,7 @@ set_target_properties(rascaline::shared PROPERTIES
     # since it might contains chemfiles
     IMPORTED_LINK_INTERFACE_LANGUAGES CXX
 )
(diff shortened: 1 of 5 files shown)


Luthaf:
The first commit update our compilation target from C++11 to C++17, which should be supported by every compiler we care about (Debian stable, ubuntu 20.04, macOS, MSVC >2019, …)

Luthaf:
> I find the name selected_keys a bit confusing, since this is a set of keys that can also extend the descriptor while the word "select" has this connotation of filtering, but for that it is too late.

The same is true for `selected_samples` and `selected_properties`. I'm not against renaming these options, but I could not find a nicer name at the time

PR #208, opened 2023-07-25
Newer Python version for upper bound CI test

PicoCentauri:
Lifting the Python version in the CI to 3.11. Additionally I added a new CI job to test our minimal supported Python version.

----
:books: Documentation preview :books:: https://rascaline--208.org.readthedocs.build/en/208/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,29 +21,38 @@ jobs:
           - os: ubuntu-20.04
             rust-version: stable
             rust-target: x86_64-unknown-linux-gnu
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -21,29 +21,38 @@ jobs:
           - os: ubuntu-20.04
```
Luthaf, line 21: I think it is time to split this into multiple tests, I'll do it in a separate PR

PicoCentauri: Thought the same. If you do it I review.

PR #212, opened 2023-08-01
General spliner class

PicoCentauri:
- Replace generate_splines by a more flexible class
- Allow to calculate the derivative of the radial integral numerically
- Add [linter](https://pypi.org/project/flake8-sphinx-links/) for to check docstrings for double backticked strings which should be links to the Python documentation
- Some minor cleanup to the rascaline `__init__.py`, i.e. adding an `__all__` list.

Splitting the code into a `SplinerBase` and `SplinerFromFunction` class sounds maybe like an overkill here. However, this is only the first commit and should serve as a discussion basis for the architecture of the classes. After we agree on the design I will use the base class to extend the splining classes to more real and also kspace splining classes.

----
:books: Documentation preview :books:: https://rascaline--212.org.readthedocs.build/en/212/


python/rascaline/rascaline/utils/splines.py
```diff
@@ -0,0 +1,371 @@
...
+        """Method calculating the derivatice of the radial integral."""
+
+        displacement = 1e-6
+
+        radial_integral_pos = self._radial_integral(
+            n, ell, positions + displacement / 2
+        )
+        radial_integral_neg = self._radial_integral(
+            n, ell, positions - displacement / 2
+        )
```
frostedoyster, line 73: In my experience, this can cause problems when evaluating derivatives at the edges of the domain (i.e., at 0 and r_cut) because the function might not be defined outside of them. However, you could also make the case that it's up to the user to provide a custom derivative function in such a situation.

frostedoyster (commented):
The design looks good to me, although the derived class isn't doing much additional work. Maybe we could ask for a third person's opinion on this

python/rascaline/tests/utils/splines.py
```diff
@@ -57,15 +61,18 @@ def test_splines_with_n_spline_points():
...
+    spliner = SplinerFromFunction(
+        radial_integral=sine,
+        cutoff=cutoff,
         max_radial=12,
         max_angular=9,
-        cutoff_radius=cutoff_radius,
+        radial_integral_derivative=cosine,
     )
 
+    spline_points = spliner.compute()["TabulatedRadialIntegral"]["points"]
```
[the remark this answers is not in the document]
PicoCentauri: You mean in the test or in general if we should change the interface? I think the approach with 

```python
    >>> calculator = SoapPowerSpectrum(
    ...     cutoff=8.0,
    ...     max_radial=12,
    ...     max_angular=9,
    ...     center_atom_weight=1.0,
    ...     radial_basis=spliner.compute(),
    ...     atomic_gaussian_width=1.0,  # ignored
    ...     cutoff_function={"Step": {}},
    ... )
```

is quite handy.

python/rascaline/rascaline/utils/splines.py
PicoCentauri: Yes, I agree. I think the numerical derivative is really the fallback solution if you can not provide a analytical derivative. We can clarify this a bit more in the doc string.

We can also improve the code here.

PicoCentauri:
We have to repbase again metatensor but otherwise should be fine.

python/rascaline/rascaline/utils/__init__.py
```diff
@@ -3,9 +3,10 @@
 import metatensor
 
 from .power_spectrum import PowerSpectrum
+from .splines import SplinerBase, SplinerFromFunction
 
 
-__all__ = ["PowerSpectrum"]
+__all__ = ["PowerSpectrum", "SplinerBase", "SplinerFromFunction"]
```
Luthaf, line 9:
```suggestion
```

I don't think `__all__` is really useful here.

python/rascaline/rascaline/utils/splines.py
```diff
@@ -0,0 +1,396 @@
...
+    @property
+    def _center_contribution(self) -> Union[None, np.ndarray]:
+        r"""Contribution of the central atom required for LODE calculations.
+
+        The `center_contribution` is defined as :math:`c_n = \sqrt{4π} \int dr r^2
+        R_n(r) g(r)` where :math:`g(r)` is the radially symmetric density function,
+        `R_n(r)` the radial basis function and :math:`n` the current radial channel.
+        Note that the integration range was deliberately left ambiguous since it depends
+        on the radial basis, i.e. for the GTO basis, :math:`r \in R^+` is used, while
+        :math:`r \in [0, cutoff]` for the monomial basis.
```
Luthaf, line 65: I still disagree with this formulation ^_^. this is always an integral to +\infty, but sometime the function to integrate is 0 outside of the cutoff, so the integral can be simplified. But we can always define it as an integral from 0 to +\infty.

python/rascaline/rascaline/utils/splines.py
```diff
@@ -0,0 +1,396 @@
...
+        This should be implemented in the child classes."""
+
+        return None
+
+    def _radial_integral_derivative(
+        self, n: int, ell: int, positions: np.ndarray
+    ) -> np.ndarray:
+        """Method calculating the derivatice of the radial integral."""
+
+        displacement = 1e-6
```
Luthaf, line 76: To make sure this is a valid displacement for the function, we could check that the `positions` absolute mean is at least 1

python/rascaline/rascaline/utils/splines.py
```diff
@@ -0,0 +1,396 @@
...
+
+from abc import ABC, abstractmethod
+from typing import Callable, Dict, Optional, Union
+
+import numpy as np
+
+
+class SplinerBase(ABC):
+    """Base class for splining arbitrary radial integrals.
+
```
Luthaf, line 26: Could this get a seealso pointer to the corresponding example?

python/rascaline/rascaline/utils/splines.py
```diff
@@ -0,0 +1,396 @@
...
+        radial_integral_pos = self._radial_integral(
+            n, ell, positions + displacement / 2
+        )
+        radial_integral_neg = self._radial_integral(
+            n, ell, positions - displacement / 2
+        )
+
+        return (radial_integral_pos - radial_integral_neg) / displacement
+
+    def _value_evaluator_3D(
```
Luthaf, line 87: I'm not sure I see the point of having this as a separate function. Should we inline it inside `compute` since it is only used there?

python/rascaline/rascaline/utils/splines.py
```diff
@@ -0,0 +1,396 @@
...
+        error gets below the ``accuracy`` threshold.
+    :parameter radial_integral_derivative: The derivative of the radial integral taking
+        the same paramaters as ``radial_integral``. If it is :py:obj:`None` (default),
+        finite differences are used to calculate the derivative of the radial integral.
+        It is recommended to provide this parameter if possible. Derivatives from finite
+        differences can cause problems when evaluating at the edges of the domain (i.e.,
+        at ``0`` and ``spline_cutoff``) because the function might not be defined
+        outside of the domain.
+    :parameter center_contribution: Contribution of the central atom required for LODE
+        calculations.
```
Luthaf, line 322: This should point to the corresponding property documentation in the base class

python/rascaline/examples/splined-radial-integral.py
```diff
@@ -114,16 +117,16 @@ def laplacian_eigenstate_basis_derivative(n, el, r):
...
-    max_radial,
-    max_angular,
-    cutoff,
-    requested_accuracy=1e-5,
+spliner = SplinerFromFunction(
+    radial_integral=laplacian_eigenstate_basis,
+    spline_cutoff=cutoff,
+    max_radial=max_radial,
+    max_angular=max_angular,
+    radial_integral_derivative=laplacian_eigenstate_basis_derivative,
```
Luthaf, line 128:
```suggestion
    radial_integral=laplacian_eigenstate_basis,
    radial_integral_derivative=laplacian_eigenstate_basis_derivative,
    spline_cutoff=cutoff,
    max_radial=max_radial,
    max_angular=max_angular,
```

this feels like a more logical grouping to me

python/rascaline/rascaline/utils/splines.py
```diff
@@ -0,0 +1,396 @@
...
+    :show-inheritance:
+"""
+
+from abc import ABC, abstractmethod
+from typing import Callable, Dict, Optional, Union
+
+import numpy as np
+
+
+class SplinerBase(ABC):
```
Luthaf, line 24: since this is specifically for tabulated radial integral, I think it should be reflected somewhere in the name:
```suggestion
class RadialIntegralSplinerBase(ABC):
```

python/rascaline/rascaline/utils/splines.py
```diff
@@ -0,0 +1,396 @@
...
+        h11 = h11.reshape(new_shape)
+
+        interpolated_values = (
+            h00 * p_k + h10 * delta_x * m_k + h01 * p_k_1 + h11 * delta_x * m_k_1
+        )
+
+        return interpolated_values
+
+
+class SplinerFromFunction(SplinerBase):
```
Luthaf, line 297:
```suggestion
class RadialIntegralFromFunction(SplinerBase):
```

python/rascaline/rascaline/utils/splines.py
```diff
@@ -0,0 +1,396 @@
...
+    """Compute the radial integral spline points based on a provided function.
+
+    :parameter radial_integral: Function to compute the radial integral. Function must
+        take ``n``, ``l``, and ``positions`` as inputs, where ``n`` and ``l`` are
+        integers and ``positions`` is a numpy 1-D array that contains the spline points
+        at which the radial integral will be evaluated. The function must return a numpy
+        1-D array containing the values of the radial integral. :parameter cutoff:
+        Spherical cutoff for the radial basis :parameter max_angular: number of radial
+        components
```
Luthaf, line 306:
```suggestion
        1-D array containing the values of the radial integral. 
        :parameter cutoff: Spherical cutoff for the radial basis 
```

Luthaf (commented):
looks good, I have some comments on minor API points & naming

python/rascaline/rascaline/utils/splines.py
PicoCentauri: fine for me to change it.

python/rascaline/rascaline/utils/splines.py
PicoCentauri: yes good idea!

python/rascaline/rascaline/utils/splines.py
PicoCentauri: we at least use it twice for the values and the derivatives of the splines. I think it is fine.

python/rascaline/rascaline/utils/splines.py
PicoCentauri: The properties are private. I don't know what we should point to.

python/rascaline/rascaline/utils/__init__.py
PicoCentauri: yeah it is not top level...

python/rascaline/rascaline/utils/splines.py
PicoCentauri: I think for the example it is not so interesting here. We can do it for the `RadialIntegralFromFunction`?

python/rascaline/rascaline/utils/splines.py
Luthaf: not really: this is two function disguised as one (the bool parameter switches between the functions). If you have strong feelings here, I can live with the code as is though

python/rascaline/rascaline/utils/splines.py
```diff
@@ -0,0 +1,393 @@
...
+        return None
+
+    def _radial_integral_derivative(
+        self, n: int, ell: int, positions: np.ndarray
+    ) -> np.ndarray:
+        """Method calculating the derivatice of the radial integral."""
+
+        # The hardcoded displacement is a only a valid choice of the positions are in a
+        # reasonable range.
+        assert np.abs(positions).mean() >= 1.0
```
PicoCentauri, line 74: Should we do a propper raise here? Probably it is better...

python/rascaline/rascaline/utils/splines.py
PicoCentauri: thanks. Always these issues with the auto formatting.

python/rascaline/rascaline/utils/splines.py
Luthaf: that works!

python/rascaline/rascaline/utils/splines.py
Luthaf: would be nice yes =)

python/rascaline/rascaline/utils/splines.py
```diff
@@ -0,0 +1,403 @@
+"""
+Splined radial integrals
+========================
+
+Classes for generating spline points which can be used as input for the for the radial
+integral hyper paramater option. For an complete example for the LE how to use the
+classes see :ref:`example-splines`.
```
Luthaf, line 7:
```suggestion
Classes for generating spline points which can be used as input for the radial
integral hyper parameter option. For an complete example on how to use the
classes see :ref:`example-splines`.
```

python/rascaline/rascaline/utils/splines.py
```diff
@@ -0,0 +1,403 @@
...
+
+.. autoclass:: rascaline.utils.RadialIntegralSplinerBase
+    :members:
+    :show-inheritance:
+
+.. autoclass:: rascaline.utils.RadialIntegralFromFunction
```
Luthaf, line 13: we should pick a place for these. I see you like to have them in the module docstring, and I prefer them in the docs sources. Let's chat about it in the meeting later today?

python/rascaline/rascaline/utils/splines.py
```diff
@@ -0,0 +1,403 @@
...
+        error gets below the ``accuracy`` threshold.
+    :parameter center_contribution: Contribution of the central atom required for LODE
+        calculations. The ``center_contribution`` is defined as
+
+        .. math::
+           c_n = \sqrt{4π}\int_0^\infty dr r^2 R_n(r) g(r)
+
+        where :math:`g(r)` is the radially symmetric density function, `R_n(r)` the
+        radial basis function and :math:`n` the current radial channel. This should be
+        implemented in the child classes.
```
Luthaf, line 325:
```suggestion
        radial basis function and :math:`n` the current radial channel. This should be
        precomputed and provided separately.
```

Luthaf (approved):
Looks good, once CI is happy we can merge!

Issue #213, opened 2023-08-04
`PowerSpectrum` TensorMap can not be fully merged in a single block


PicoCentauri:
When computing a descriptor with `PowerSpectrum` from two frames that have different atomtypes i.e one with only an `H` and the other one only with an `O` the result can not be densified.

Consider the following example

```python
import ase
import numpy as np
from equistore import EquistoreError
from rascaline import SoapPowerSpectrum, SphericalExpansion
from rascaline.utils import PowerSpectrum

frames = [
    ase.Atoms("H", positions=np.zeros([1, 3])),
    ase.Atoms("O", positions=np.zeros([1, 3])),
]

hypers = dict(
    cutoff=3.0,
    max_radial=6,
    max_angular=1,
    atomic_gaussian_width=0.3,
    center_atom_weight=1.0,
    radial_basis={"Gto": {}},
    cutoff_function={"ShiftedCosine": {"width": 0.5}},
)

calculator_pyps = PowerSpectrum(SphericalExpansion(**hypers))
descriptor_pyps = calculator_pyps.compute(frames)
print(descriptor_pyps)

try:
    descriptor_pyps.keys_to_samples("species_center")
except EquistoreError as err:
    print(err)
```

which prints

```
TensorMap with 2 blocks
keys: species_center
            1
            8
invalid parameter: can not move keys to samples if the blocks have different property labels
```

The same happens when calling the `SOAPPowerSpectrum`

```python
calculator_ps = SoapPowerSpectrum(**hypers)
descriptor_ps = calculator_ps.compute(frames)
print(descriptor_ps)

descriptor_moved = descriptor_ps.keys_to_properties(
    ["species_neighbor_1", "species_neighbor_2"]
)
print(descriptor_moved)

try:
    descriptor_moved.keys_to_samples("species_center")
except EquistoreError as err:
    print(err)
```

leading to

```
TensorMap with 2 blocks
keys: species_center  species_neighbor_1  species_neighbor_2
            1                 1                   1
            8                 8                   8
TensorMap with 2 blocks
keys: species_center
            1
            8
invalid parameter: can not move keys to samples if the blocks have different property labels
```

For `SOAPPowerSpectrum` the error can be solved by changing the order of `keys_to_properties` and `keys_to_samples`. For `PowerSpectrum` we can not do this because we already move the keys to properties within the computation to avoid nasty bookkeeping. However, it might be that we have to do this because I see no current way to solve this.

Somebody has ideas?

[...]

PR #214, opened 2023-08-07
Add option to fill all species  when computing `PowerSpectrum` 

PicoCentauri:
Possible fix for #213.

This adds the option `fill_species_neighbor ` to the compute method of  `PowerSpectrum`. Setting this to `True` changes `keys_to_move` when performing `keys_to_properties` in such a way that it contains all species of the provided systems.

I am not super happy with the name of the parameter but I think the it should be part of the the compute method.

----
:books: Documentation preview :books:: https://rascaline--214.org.readthedocs.build/en/214/

### DIFF
--- a/python/rascaline/rascaline/utils/power_spectrum.py
+++ b/python/rascaline/rascaline/utils/power_spectrum.py
@@ -183,9 +183,17 @@ class PowerSpectrum:
         ]
         assert spherical_expansion_1.keys.names == expected_key_names
         assert spherical_expansion_1.property_names == ["n"]
-        spherical_expansion_1 = spherical_expansion_1.keys_to_properties(
-            "species_neighbor"
+
+        # Fill blocks with `species_neighbor` from ALL blocks. If we don't do this
+        # merging blocks along the ``sample`` direction might be not possible.
+        keys_to_move = Labels(
+            names="species_neighbor",
+            values=np.unique(spherical_expansion_1.keys["species_neighbor"]).reshape(
+                -1, 1
+            ),
         )
+
+        spherical_expansion_1 = spherical_expansion_1.keys_to_properties(keys_to_move)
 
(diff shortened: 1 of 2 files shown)


ceriottm:
I think the ideal implementation would be something that has an option `neighbor_species` that defaults to None (and then automatically fetches them from the inputs) or a list of labels or numbers (and then makes sure that all the missing ones are padded with zeros.

python/rascaline/rascaline/utils/power_spectrum.py
```diff
@@ -183,9 +190,18 @@ def compute(
...
-            "species_neighbor"
-        )
+
+        if fill_species_neighbor:
+            keys_to_move = Labels(
+                names="species_neighbor",
+                values=np.unique(
+                    spherical_expansion_1.keys["species_neighbor"]
+                ).reshape(-1, 1),
+            )
```
Luthaf, line 200: I think I'd be fine with this being the only behavior: we are not introducing too much zeros since we are moving the two spherical expansions separately, so we could just always do this and not have an extra option.

Luthaf:
> I think the ideal implementation would be something that has an option neighbor_species that defaults to None (and then automatically fetches them from the inputs) or a list of labels or numbers (and then makes sure that all the missing ones are padded with zeros.

That could work as well, but I don't love that we are introducing a new set of parameters just for this. This feels very similar to the job done by `selected_properties`. Implementing full support for `selected_properties` in Python would be relatively hard, but we should also consider it!

> I am not super happy with the name of the parameter but I think the it should be part of the the compute method.

The issue with making it part of the compute method is that we now have different set of options for Rust calculators and this Python calculator, which I don't love. I agree it feels much more natural to have it here though!

Luthaf (approved):

PR #216, opened 2023-08-14
Allow cell matrix to have a negative determinant

### LINKED ISSUE #215: `RascalError: internal error: matrix is not invertible` for Si decohesion frames.
Using the dataset from Gabor Csanyi et al., DOI: https://doi.org/10.1103/physrevx.8.041048, for some of the decohesion frames, I get the above rascal error when computing the SphericalExpansion.

The original raw data can be obtained from https://www.repository.cam.ac.uk/items/1cfce189-65c3-4e05-a430-ba062a787802. With respect to the total set of frames from gp_iter6_sparse9k.xml.xyz, the indices of the offending frames are `[2159, 2160, 2161, 2162, 2163, 2164, 2165, 2166, 2167, 2168, 2169]`, and I have attached these in an xyz.

To narrow down the problem: it appears that in the offending frames, but not in the rest, the Z-cell vector has a single large negative component. Any ideas as to why rascaline doesn't like this?

Tagging @SanggyuChong as we experienced this problem together.

[failed.xyz.zip](https://github.com/Luthaf/rascaline/files/12319958/failed.xyz.zip)

Luthaf:
Fix #215

Disallowing cells with a negative determinant seems to have been an oversight from me. I checked the neighbours list rascaline computes for all the structures in #215, and they all match the neighbor lists computed with ASE, so I think everything is fine? 

If someone knows more about this kind of cell, feel free to chime in! I've never seen such cells before, but they look legit.

----
:books: Documentation preview :books:: https://rascaline--216.org.readthedocs.build/en/216/

### DIFF
--- a/rascaline/src/systems/cell.rs
+++ b/rascaline/src/systems/cell.rs
@@ -39,7 +39,7 @@ impl From<Matrix3> for UnitCell {
             return UnitCell::infinite()
         }
 
-        assert!(matrix.determinant() > 1e-6, "matrix is not invertible");
+        assert!(matrix.determinant().abs() > 1e-6, "matrix is not invertible");
 
         let is_close_0 = |value| f64::abs(value) < 1e-6;
         let is_diagonal = |matrix: Matrix3| {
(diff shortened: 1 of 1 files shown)


Luthaf:
Ping @jwa7 @SanggyuChong

ceriottm:
You can always make the determinant negative by swapping two axes. Many codes enforce special conditions on the cell (LAMMPS for instance does) to make them as "compact" as possible, and I think that in the process they also make the determinant positive. Not strictly necessary though IMO

SanggyuChong:
Thanks for this, @Luthaf! Changes look good to me.

jwa7 (approved):
Thanks for the quick fix, looks good @Luthaf :)

Issue #18, opened 2021-03-29
Add tutorial on "implementing a new system type"


Luthaf:
This would be an advanced tutorial, and could even live in the developer documentation

PR #217, opened 2023-08-16
Add an example of how to use SystemBase

Luthaf:
Related to #18, only for Python in this case.

----
:books: Documentation preview :books:: https://rascaline--217.org.readthedocs.build/en/217/


PicoCentauri:
I am happy but the linter not 🥺

PR #220, opened 2023-08-17
Make Python PS torch compatible

PicoCentauri:
----
:books: Documentation preview :books:: https://rascaline--220.org.readthedocs.build/en/220/

### DIFF
--- a/python/rascaline-torch/rascaline/torch/calculator_base.py
+++ b/python/rascaline-torch/rascaline/torch/calculator_base.py
@@ -57,13 +57,20 @@ class CalculatorModule(torch.nn.Module):
 
     def __init__(self, name: str, parameters: str):
         """"""
+        # empty docstring here for the docs to render corectly
         super().__init__()
+        self._c_name = name
         self._c = CalculatorHolder(name=name, parameters=parameters)
 
     @property
     def name(self) -> str:
         """name of this calculator"""
         return self._c.name
+
+    @property
+    def c_name(self) -> str:
+        """name used to register & create this calculator"""
+        return self._c_name
 
     @property
     def parameters(self) -> str:
@@ -79,6 +86,7 @@ class CalculatorModule(torch.nn.Module):
         self,
         systems: Union[System, List[System]],
         gradients: Optional[List[str]] = None,
+        use_native_system: bool = True,
     ) -> equistore.torch.TensorMap:
         """Runs a calculation with this calculator on the given ``systems``.
 
@@ -97,6 +105,10 @@ class CalculatorModule(torch.nn.Module):
             ``None`` or an empty list ``[]``, no gradients are kept in the output. Some
             gradients might still be computed at runtime to allow for backward
             propagation.
+
+        :param use_native_system: This can only be :py:obj:`True`, and is here for
+            compatibility with the same parameter on
+            :py:meth:`rascaline.calculators.CalculatorBase.compute`.
         """
         if gradients is None:
             gradients = []
@@ -104,13 +116,20 @@ class CalculatorModule(torch.nn.Module):
         if not isinstance(systems, list):
             systems = [systems]
 
+        # We have this parameter to have the same API as rascaline.
+        if not use_native_system:
+            raise ValueError("only `use_native_system=True` is supported")
+
         return self._c.compute(systems=systems, gradients=gradients)
 
     def forward(
         self,
         systems: List[System],
         gradients: Optional[List[str]] = None,
+        use_native_system: bool = True,
     ) -> equistore.torch.TensorMap:
         """forward just calls :py:meth:`CalculatorModule.compute`"""
 
-        return self.compute(systems=systems, gradients=gradients)
+        return self.compute(
+            systems=systems, gradients=gradients, use_native_system=use_native_system
+        )
--- a/python/rascaline-torch/rascaline/torch/utils/power_spectrum.py
+++ b/python/rascaline-torch/rascaline/torch/utils/power_spectrum.py
@@ -1,0 +1,94 @@
+import importlib
+import sys
+from typing import List, Optional, Union
+
+import torch
+from equistore.torch import Labels, TensorBlock, TensorMap
+
+import rascaline.utils.power_spectrum
+
+from ..calculator_base import CalculatorModule as CalculatorBase
+from ..system import System as IntoSystem
+
+
+# For details what is happening here take a look an `rascaline.torch.calculators`.
+
+# Step 1: create te `_classes` module as an empty module
+spec = importlib.util.spec_from_loader(
+    "rascaline.torch.utils.power_spectrum._classes",
+    loader=None,
+)
+module = importlib.util.module_from_spec(spec)
+# This module only exposes a handful of things, defined here. Any changes here MUST also
+# be made to the `equistore/operations/_classes.py` file, which is used in non
+# TorchScript mode.
+module.__dict__["Labels"] = Labels
+module.__dict__["TensorBlock"] = TensorBlock
+module.__dict__["TensorMap"] = TensorMap
+module.__dict__["CalculatorBase"] = CalculatorBase
+module.__dict__["IntoSystem"] = IntoSystem
+
+# register the module in sys.modules, so future import find it directly
+sys.modules[spec.name] = module
+
+
+# Step 2: create a module named `rascaline.torch.utils.power_spectrum` using code from
+# `rascaline.utils.power_spectrum`
+spec = importlib.util.spec_from_file_location(
+    "rascaline.torch.utils.power_spectrum",
+    rascaline.utils.power_spectrum.__file__,
+)
+
+module = importlib.util.module_from_spec(spec)
+sys.modules[spec.name] = module
+spec.loader.exec_module(module)
+
+
+# Store the original class to avoid recursion problems
+PowerSpectrumBase = module.PowerSpectrum
+
+
+class PowerSpectrum(torch.nn.Module, PowerSpectrumBase):
+    """
+    Torch version of the general power spectrum of one or of two calculators.
+
+    The class provides :py:meth:`PowerSpectrum.forward` and the integration with
+    :py:class:`torch.nn.Module`. For more details see
+    :py:class:`rascaline.utils.PowerSpectrum`.
+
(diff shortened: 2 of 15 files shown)


python/rascaline/rascaline/utils/_helpers.py
```diff
@@ -0,0 +1,8 @@
+from rascaline.utils import _dispatch
```
PicoCentauri, line 1: @Luthaf I am not super happy with the name. You have a better idea how we can organize the `_dispatch`

Luthaf: with the `_dispatch` name?

python/rascaline-torch/rascaline/torch/calculator_base.py
```diff
@@ -56,15 +56,20 @@ class CalculatorModule(torch.nn.Module):
     """
 
     def __init__(self, name: str, parameters: str):
-        """"""
```
Luthaf, line 59: I remember we need the empty docstring here for the docs to render right, I'll check again

python/rascaline/rascaline/utils/_helpers.py
Luthaf: Also, this is different from what we do in equistore.torch.operations: over there the `_classes` module is only used to define the classes, `_dispatch` is imported as usual. 

The trick there is to start by creating a custom `equistore.torch.operations._classes` module, and registering it in `sys.module`. Then when importing the rest of the code, `from ._classes import XXX`  resolves to this custom module without re-importing the file.

python/rascaline/rascaline/utils/power_spectrum.py
```diff
@@ -2,14 +2,34 @@
...
+
+try:
+    _ = IntoSystem
+except NameError:
+    from ..systems import IntoSystem
+
+try:
+    _ = torch_nn_module
+except NameError:
+    torch_nn_module = object
```
Luthaf, line 29: All this should be removed and replaced by `from ._classes import Labels, TensorBlock, TensorMap, IntoSystem, CalculatorBase, torch_nn_module`. And then in `_classes.py` these are the pure Python versions of the classes; and in the manually created `rascaline.torch.utils._classes` module these are the TorchScript version of the classes.

python/rascaline-torch/rascaline/torch/calculator_base.py
```diff
@@ -97,20 +103,32 @@ def compute(
             ``None`` or an empty list ``[]``, no gradients are kept in the output. Some
             gradients might still be computed at runtime to allow for backward
             propagation.
+
+        :param use_native_system: If :py:obj:`True` (this is the default), copy data
+            from the ``systems`` into Rust ``SimpleSystem``. This can be a lot
+            faster than having to cross the FFI boundary often when accessing
+            the neighbor list. :py:obj:`False` is not supported.
```
Luthaf, line 110:
```suggestion
        :param use_native_system: This can only be :py:obj:`True`, and is here for compatibility 
            with the same parameter on :py:meth:`rascaline.calculators.CalculatorBase.compute`.
```

python/rascaline-torch/rascaline/torch/utils/power_spectrum.py
```diff
@@ -0,0 +1,52 @@
...
+import rascaline.utils.power_spectrum
+from rascaline.utils import _dispatch
+
+from ..calculator_base import CalculatorModule
+from ..system import System
+
+
+# For details what is happening here take a look an `rascaline.torch.calculators`.
+
+# Step 1: create te `_helpers` module as an empty module
```
Luthaf, line 16:
```suggestion
# Step 1: create the `_helpers` module as an empty module
```

python/rascaline-torch/tests/utils/power_spectrum.py
```diff
@@ -0,0 +1,51 @@
...
+
+def check_operation(PowerSpectrum):
+    # this only runs basic checks functionality checks, and that the code produces
+    # output with the right type
+
+    descriptor = calculator().compute(system(), gradients=["positions"])
+
+    assert isinstance(descriptor, torch.ScriptObject)
+    if version.parse(torch.__version__) >= version.parse("2.1"):
+        assert descriptor._type().name() == "TensorMap"
```
Luthaf, line 41: this test is not using the PowerSpectrum class. 

```suggestion
def check_operation(calculator):
    # this only runs basic checks functionality checks, and that the code produces
    # output with the right type

    descriptor = calculator.compute(system(), gradients=["positions"])

    assert isinstance(descriptor, torch.ScriptObject)
    if version.parse(torch.__version__) >= version.parse("2.1"):
        assert descriptor._type().name() == "TensorMap"
```

and rename `def calculator()` to `def spherical_expansion_calculator()`

python/rascaline/rascaline/utils/_classes.py
```diff
@@ -0,0 +1,20 @@
+from equistore.core import Labels, TensorBlock, TensorMap
+
+from ..calculator_base import CalculatorBase
+from ..systems import IntoSystem
+from . import _dispatch
```
Luthaf, line 5: Do you want to change this before merging?

python/rascaline-torch/tests/utils/power_spectrum.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def test_operation_as_python():
+    check_operation(PowerSpectrum)
+
+
+def test_operation_as_torch_script():
+    scripted = torch.jit.script(calculator())
```
Luthaf, line 49:
```suggestion
    calculator = PowerSpectrum(spherical_expansion_calculator())
    scripted = torch.jit.script(calculator)
```

python/rascaline/rascaline/utils/_classes.py
Luthaf: instead of making `_dispatch` a part of `_classes`, just import it `from . import _dispatch` in the code, like we do for the operations

python/rascaline/rascaline/utils/power_spectrum/calculator.py
```diff
@@ -287,23 +312,28 @@ def _positions_gradients(new_block, block_1, block_2, factor):
...
-        gradient_values = np.zeros(
-            [gradients_samples.values.shape[0], 3, len(new_block.properties)]
+        gradient_values = _dispatch.zeros_like(
+            array=gradient_1.values,
+            shape=[gradients_samples.values.shape[0], 3, len(new_block.properties)],
         )
 
         # this is equivalent to an einsum with: ixma, imb -> ixab
-        gradient_1_values = factor * np.matmul(
+        block_2_values = block_2.values[gradient_1.samples.column("sample")]
```
Luthaf, line 321: you might have to do the same as in https://github.com/lab-cosmo/equistore/pull/332 with `_dispatch.to_index_array` here. This is only an issue on older torch version

PicoCentauri: Ohh yes thanks!

python/rascaline-torch/rascaline/torch/utils/power_spectrum.py
```diff
@@ -0,0 +1,105 @@
...
+
+# # TODO: explain
+# spec = importlib.util.spec_from_file_location(
+#     "rascaline.torch.utils._dispatch",
+#     rascaline.utils._dispatch.__file__,
+# )
+# module = importlib.util.module_from_spec(spec)
+# sys.modules[spec.name] = module
+# spec.loader.exec_module(module)
+
```
Luthaf, line 44:
```suggestion
```

python/rascaline-torch/rascaline/torch/utils/power_spectrum.py
```diff
@@ -0,0 +1,105 @@
...
+    "rascaline.torch.utils.power_spectrum",
+    rascaline.utils.power_spectrum.__file__,
+)
+
+module = importlib.util.module_from_spec(spec)
+sys.modules[spec.name] = module
+spec.loader.exec_module(module)
+
+
+# Define the class to avoid lookup and recursion problems
```
Luthaf, line 58:
```suggestion
# Store the original class to avoid recursion problems
```

python/rascaline-torch/rascaline/torch/utils/power_spectrum.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def forward(
+        self,
+        systems: Union[IntoSystem, List[IntoSystem]],
+        gradients: Optional[List[str]] = None,
+        use_native_system: bool = True,
+    ) -> TensorMap:
+        """forward just calls :py:meth:`PowerSpectrum.compute`"""
+
+        return self.compute(
```
Luthaf, line 98: should we do something like this to avoid issues with a possible `compute` method on torch.nn.Module?

```suggestion
        return PowerSpectrumBase.compute(self,
```

Luthaf (commented):
A bit more cleanup and this should be good to go!

PicoCentauri:
Thanks for the cleanup!

python/rascaline-torch/rascaline/torch/utils/power_spectrum.py
PicoCentauri: Seems like torch doesn't like this.

PicoCentauri:
I wonder why torch was complaining about the change of the `forward` function. The compiler was telling

```
E           Unknown type name 'CalculatorBase':
E             File "/Users/philiploche/repos/rascaline/.tox/torch-tests/lib/python3.11/site-packages/rascaline/utils/power_spectrum/calculator.py", line 12
E               def __init__(
E                   self,
E                   calculator_1: CalculatorBase,
E                                 ~~~~~~~~~~~~~~ <--- HERE
E                   calculator_2: Optional[CalculatorBase] = None,
E               ):
E           'PowerSpectrum.__init__' is being compiled since it was called from '__torch__.rascaline.torch.utils.power_spectrum.calculator.PowerSpectrum'
```

But he already had compiled the non-torch Power Spectrum version...

Luthaf:
This also required some changes in time-graph to make valgrind happy: https://github.com/Luthaf/time-graph/commit/eabad483c2454e8d5a36e70075a1dff0b83d83e8

PR #221, opened 2023-08-21
Replace pair_id in neighbors list samples with the actual cell shift

Luthaf:
This is more informative, and not that much work overall.

@DivyaSuman14 and @HannaTuerk needed something like this for the hamiltonian learning. You can even compute only one of the translated cells at the time by using sample selection!

---

@felixmusil, this should solve https://github.com/Luthaf/rascaline/pull/131#issuecomment-1332201981, with all the data you want being in the samples:

```py
calculator = rascaline.NeighborList(cutoff=3.4, full_neighbor_list=True)
descriptor = calculator.compute(frames)
descriptor.block(0).samples
# Labels(
#     structure  first_atom  second_atom  cell_shift_a  cell_shift_b  cell_shift_c
#         0          0           37            0             0             0
#         0          37           0            0             0             0
#         0          1            2            0             0             0
# ...
#         0          51          51            0             0             0
#         0          52          52            0             0             0
#         0          53          53            0             0             0
# )

# in ASE terms:
i = descriptor.block(0).samples["first_atom"]
j = descriptor.block(0).samples["second_atom"]
S = descriptor.block(0).samples.view(["cell_shift_a", "cell_shift_b", "cell_shift_c"]).values
D = descriptor.block(0).values
```

---

Ping @rosecers I think you are also using the neighbors list calculator so these changes might impact you (only if you are explicitly manipulating the old `pair_id` sample)

----
:books: Documentation preview :books:: https://rascaline--221.org.readthedocs.build/en/221/


felixmusil:
Looks perfect, great work !

rosecers (commented):
Nice! @arthur-lin1027 can you check if this makes any problems with anisoap?

arthur-lin1027:
> Nice! @arthur-lin1027 can you check if this makes any problems with anisoap?

I ran the code on `dev424` (my locally installed version) and `dev450` (this current version) and I got the same results! Makes sense since we aren't manipulating the `pair_id` sample, as Guillaume mentioned above.

python/rascaline/rascaline/calculators.py
```diff
@@ -37,33 +37,54 @@ def __init__(self, cutoff, delta, name):
...
+        cutoff,
+        full_neighbor_list,
+        self_pairs=False,
+        quantity="Distance",
+    ):
         parameters = {
             "cutoff": cutoff,
             "full_neighbor_list": full_neighbor_list,
             "self_pairs": self_pairs,
+            "quantity": quantity,
```
PicoCentauri, line 87: I don't really see where `quantity` is also introduced in the rust code. Is this parameter already there?

python/rascaline/rascaline/calculators.py
```diff
@@ -37,33 +37,54 @@ def __init__(self, cutoff, delta, name):
...
+    pairs between the same atom (if the cutoff is larger than the cell).
     """
 
-    def __init__(self, cutoff, full_neighbor_list, self_pairs=False):
+    def __init__(
+        self,
+        cutoff,
+        full_neighbor_list,
+        self_pairs=False,
+        quantity="Distance",
```
PicoCentauri, line 81: Maybe some typehints since you are already touching these functions.

python/rascaline/rascaline/calculators.py
Luthaf: this is a leftover from another solution we did not use in the end, good catch!

PicoCentauri (approved):
Thanks for incoperating the comments. LGTM!

PR #234, opened 2023-09-21
Add coverage for Python torch bindings

PicoCentauri:
Using the same logic and structure as https://github.com/lab-cosmo/metatensor/pull/381.

----
:books: Documentation preview :books:: https://rascaline--234.org.readthedocs.build/en/234/

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -38,7 +38,7 @@ jobs:
       - name: install python dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install tox
+          python -m pip install tox coverage
 
       - name: install lcov
         run: sudo apt install -y lcov
@@ -47,7 +47,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: .tox
-          key: tox-${{ matrix.os }}-${{ hashFiles('pyproject.toml', 'setup.cfg', 'tox.ini') }}
+          key: tox-${{ hashFiles('pyproject.toml', 'setup.cfg', 'tox.ini') }}
 
       - name: Setup sccache
         uses: mozilla-actions/sccache-action@v0.0.3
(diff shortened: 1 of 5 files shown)


Luthaf:
It looks like the coverage is very confused... it is reporting empty lines as not-covered: https://app.codecov.io/github/Luthaf/rascaline/commit/ae4cb69ff05f267a229737158a11581baafc7e56/blob/python/rascaline-torch/rascaline/torch/__init__.py, just above `_load_library()`

.github/workflows/coverage.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: .tox
-          key: tox-${{ matrix.os }}-${{ hashFiles('pyproject.toml', 'setup.cfg', 'tox.ini') }}
+          key: tox-${{ hashFiles('pyproject.toml', 'setup.cfg', 'tox.ini') }}
```
Luthaf, line 50: why remove matrix.os from the cache key?

tox.ini
```diff
@@ -183,3 +186,19 @@ commands =
...
+
+[coverage:report]
+skip_covered = True
+show_missing = True
+omit =
+    tests/.*
+    examples/.*
+
+[coverage:paths]
+notorch =
```
Luthaf, line 198:
```suggestion
rascaline =
```

tox.ini
```diff
@@ -183,3 +186,19 @@ commands =
...
+    python/rascaline/rascaline
+    .tox/*/lib/python*/site-packages/rascaline
+
+torch =
```
Luthaf, line 202:
```suggestion
rascaline_torch =
```

Luthaf (commented):
looks good overall!

.github/workflows/coverage.yml
PicoCentauri: Because, we have no matrix here. We only run this on ubuntu 22

Luthaf:
Seems that codecov is no longer adding a commit check with the overall coverage, we'll have to look into this. But this PR should be fine!

PR #235, opened 2023-09-22
Added real and Fourier space spliner base classes

PicoCentauri:
As given in the title this PR adds two new base clases:

1. `RealSpaceRadialIntegralSplinerBase`
2. `FourierSpaceRadialIntegralSplinerBase`

With these base classes constructing (almost) any radial integral becomes quite easy. A developer/user **only** has to **implement** the method for her/his **radial basis** in a child class. The actual integral is performed numerically with `quad`. Also I added the option to (ortho)normalize the radial integral.

For `RealSpaceRadialIntegralSplinerBase` I added support for delta atomic densities and Gaussian atomic densities. The support for custom atomic densities is prepared, but the logic is missing. Since the integral is a more complex we should do this in an upcoming PR.

`FourierSpaceRadialIntegralSplinerBase` already works for any atomic density and radial basis.

# TODO

I still have to document and test everything. Especially the equations for the radial integral should be added.

As a test I suggest that I add a GTO child class for LODE and SOAP and compare this to the analytical results which we already implemented in rust.

I would happy about some comments if the class structure makes sense and is understandable :-). 

----
:books: Documentation preview :books:: https://rascaline--235.org.readthedocs.build/en/235/

### DIFF
--- a/python/rascaline/rascaline/utils/atomic_density.py
+++ b/python/rascaline/rascaline/utils/atomic_density.py
@@ -1,0 +1,166 @@
+r"""
+.. _python-atomic-density:
+
+Atomic Density
+==============
+
+the atomic density function :math:`g(r)`, often chosen to be a Gaussian or Delta
+function, that defined the type of density under consideration. For a given center atom
+:math:`i` in the structure, the total density function :math:`\rho_i(\boldsymbol{r})`
+around is then defined as :math:`\rho_i(\boldsymbol{r}) = \sum_{j} g(\boldsymbol{r} -
+\boldsymbol{r}_{ij})`.
+
+All atomic densities are based on
+
+.. autoclass:: rascaline.utils.atomic_density.AtomicDensityBase
+    :members:
+    :show-inheritance:
+
+In addition, we provide the following explicit implementations
+
+.. autoclass:: rascaline.utils.atomic_density.DeltaDensity
+    :members:
+    :show-inheritance:
+
+.. autoclass:: rascaline.utils.atomic_density.GaussianDensity
+    :members:
+    :show-inheritance:
+
+.. autoclass:: rascaline.utils.atomic_density.LodeDensity
+    :members:
+    :show-inheritance:
+
+"""
+from abc import ABC, abstractmethod
+from typing import Union
+
+import numpy as np
+
+
+try:
+    from scipy.special import gamma, gammainc
+
+    HAS_SCIPY = True
+except ImportError:
+    HAS_SCIPY = False
+
+
+class AtomicDensityBase(ABC):
+    """Base class implemententing atomic densities."""
+
+    @abstractmethod
+    def compute(self, positions: Union[float, np.ndarray]) -> Union[float, np.ndarray]:
+        """Method calculating the atomic density.
+
+        :param positions: positions to evaluate the atomic densities
+        :returns: evaluated atomic density
+        """
+        ...
+
+
+class DeltaDensity(AtomicDensityBase):
+    r"""Delta atomic densities of the form :math:`g(r)=\delta(r)`."""
+
+    def compute(self, positions: Union[float, np.ndarray]) -> Union[float, np.ndarray]:
+        raise ValueError(
+            "Compute function of the delta density should never called directly."
+        )
+
+
+class GaussianDensity(AtomicDensityBase):
+    r"""Gaussian atomic density function.
+
+    In rascaline, we use the convention
+
+    .. math::
+
+        g(r) = \frac{1}{(\pi \sigma^2)^{3/4}}e^{-\frac{r^2}{2\sigma^2}} \,.
+
+    The prefactor was chosen such that the "L2-norm" of the Gaussian
+
+    .. math::
+
+            \|g\|^2 = \int \mathrm{d}^3\boldsymbol{r} |g(r)|^2 = 1\,,
+
+    :param atomic_gaussian_width: Width of the atom-centered gaussian used to create the
+        atomic density
+    """
+
+    def __init__(self, atomic_gaussian_width: float):
+        self.atomic_gaussian_width = atomic_gaussian_width
+
+    def compute(self, positions: Union[float, np.ndarray]) -> Union[float, np.ndarray]:
+        atomic_gaussian_width_sq = self.atomic_gaussian_width**2
+        return np.exp(-0.5 * positions**2 / atomic_gaussian_width_sq) / (
+            np.pi * atomic_gaussian_width_sq
+        ) ** (3 / 4)
+
+
+class LodeDensity(AtomicDensityBase):
+    r"""Smeared Power Law Densities/
+
+    It is defined as
+
+    .. math::
+
+        g(r) = \frac{1}{\Gamma\left(\frac{p}{2}\right)}
+               \frac{\gamma\left( \frac{p}{2}, \frac{r^2}{2\sigma^2} \right)}
+                    {r^p},
+
+    where :math:`\Gamma(z)` is the Gamma function and :math:`\gamma(a, x)` is the
+    incomplete lower Gamma function. However its evaluation at :math:`r=0` is
+    problematic because :math:`g(r)` is of the form :math:`0/0`. For practical
+    implementations, it is thus more convenient to rewrite the density as
+
+    .. math::
+
+        g(r) = \frac{1}{\Gamma(a)}\frac{1}{\left(2 \sigma^2\right)^a}
+                \begin{cases}
+                    \frac{1}{a} - \frac{x}{a+1} + \frac{x^2}{2(a+2)} + \mathcal{O}(x^3)
+                        & x < 10^{-5} \\
+                    \frac{\gamma(a,x)}{x^a}
+                        & x \geq 10^{-5}
+                \end{cases}
+
+    It is convenient to use the expression for sufficiently small :math:`x` since the
+    relative weight of the first neglected term is on the order of :math:`1/6x^3`.
+    Therefore, the threshold :math:`x = 10^{-5}` leads to relative errors on the order
+    of the machine epsilon.
+
+    :param atomic_gaussian_width: Width of the atom-centered gaussian used to create the
+        atomic density
+    :param potential_exponent: Potential exponent of the decorated atom density.
+        Currently only implemented for potential_exponent < 10. Some exponents can be
+        connected to SOAP or physics-based quantities: p=0 uses Gaussian densities as in
+        SOAP, p=1 uses 1/r Coulomb like densities, p=6 uses 1/r^6 dispersion like
+        densities.
+    """
+
+    def __init__(self, atomic_gaussian_width: float, potential_exponent: int):
+        if not HAS_SCIPY:
+            raise ValueError("LodeDensity requires scipy to be installed")
+
+        self.potential_exponent = potential_exponent
+        self.atomic_gaussian_width = atomic_gaussian_width
+
+    def _short_range(self, a, x):
+        return 1 / a - x / (a + 1) + x**2 / (2 * (a + 2))
+
+    def _long_range(self, a, x):
+        return gamma(a) * gammainc(a, x) / x**a
+
+    def compute(self, positions: Union[float, np.ndarray]) -> Union[float, np.ndarray]:
+        if self.potential_exponent == 0:
+            return GaussianDensity.compute(self, positions=positions)
+        else:
+            atomic_gaussian_width_sq = self.atomic_gaussian_width**2
+            a = self.potential_exponent / 2
+            x = positions**2 / (2 * atomic_gaussian_width_sq)
+
+            prefac = 1 / gamma(a) / (2 * atomic_gaussian_width_sq) ** a
+
+            return prefac * np.where(
+                x < 1e-5,
+                self._short_range(a, x),
+                self._long_range(a, x),
+            )
--- a/python/rascaline/rascaline/utils/radial_basis.py
+++ b/python/rascaline/rascaline/utils/radial_basis.py
@@ -1,0 +1,325 @@
+r"""
+.. _python-radial-basis:
+
+Radial Basis
+============
+
+Radial basis functions :math:`R_{nl}(\boldsymbol{r})` are besides :ref:`atomic densities
+<python-atomic-density>` :math:`\rho_i` the central ingredents to compute spherical
+expansion coefficients :math:`\langle anlm\vert\rho_i\rangle`. Radial basis functions,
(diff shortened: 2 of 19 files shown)


Luthaf:
Regarding the class structure, should we have separate classes for atomic density and basis functions instead of putting everyone inside the spliner? Something like

```py
class GaussianDensity(AtomicDensityBase):
    ...


class DeltaDensity(AtomicDensityBase):
    ...


class RadiallyVaryingDensity(AtomicDensityBase):
    ...




class GTOBasis(RadialBasis):
    ...


class MonomialBasis(RadialBasis):
    ...




class RealSpaceSpliner():
    def __init__(self, density: AtomicDensityBase, basis: RadialBasis):
        ...


class KSpaceSpliner():
    def __init__(self, density: AtomicDensityBase, basis: RadialBasis):
        ...
```

This way `RealSpaceSpliner` and `KSpaceSpliner` can be concrete classes, and the users only need to customize the part they care about (density and/or basis).

----

A partially unrelated thing I find weird is making the `abstractmethod` functions private (starting with an `_`). If they are part of the abstract interface, they should be public IMO. 

Which is also solved by the API above, using composition instead of inheritance for the density & basis functions!

PicoCentauri:
Yes, we can split this up even more. Shouldn't we then use inheritance, i.e.

```python
class GTORealSpaceSpliner(RealSpaceSpliner, GTOBasis, GaussianDensity):
    ....
```

Luthaf:
I think composition is going to be much clearer than multiple inheritance here

PicoCentauri:
Yeah might be. First of all if to fix the code that it is actually working. Currently, I run into infinite loops when doing the integrals.

PicoCentauri:
@Luthaf sorry for the delay but I continued the work on the spliner classes. Splitting the `AtomicDensities` and the `RadialBasis` was a very good idea. Looks much cleaner now. There is still a lot of docs missing, but maybe you can another look?

The LODE part is already working and agrees with the analytical implementation that we did in rust as the `test_fourier_space_spliner()` is showing. 🥳

I will continue the work on the real space part next week.

python/rascaline/rascaline/utils/splines.py
```diff
@@ -401,3 +448,195 @@ def _radial_integral_derivative(
...
+        )
+
+    def _radial_integral(self, n: int, ell: int, positions: np.ndarray) -> np.ndarray:
+        # Use analytical integrals if available
+        if type(self.density) is DeltaDensity:
+            return self._radial_integral_delta(n, ell, positions, derivative=False)
+        elif type(self.density) is GaussianDensity:
+            return self._radial_integral_gaussian(n, ell, positions, derivative=False)
+        else:
+            return self._radial_integral_custom(n, ell, positions, derivative=False)
```
PicoCentauri, line 483: One could factor `_radial_integral_delta` and `_radial_integral_gaussian` out, but I would keep this logic and the implementation here since it combines the atomic densities and radial basis.

PicoCentauri:
After some rounds of debugging the implementation is working. The main and most important tests are testing that the Numerical implementation for the special case of GTOs + Gaussian densities agrees with the analytical implementation in rust for GTOs + Gaussian densities.

In addition I already added some more interesting densities and basis classes.

docs/src/explanations/radial-integral.rst
```diff
@@ -0,0 +1,278 @@
...
+which are completely specified by
+
+-  the density function :math:`g(\boldsymbol{r})`
+
+-  the radial basis :math:`R_{nl}(r)`
+
+-  the position of the neighbor atom :math:`\boldsymbol{r}_{ij}` relative to the center
+   atom
+
+The Radial Integral Class
```
Luthaf, line 84: I would rather use normal case in titles:
```suggestion
The radial integral class
```

docs/src/explanations/radial-integral.rst
```diff
@@ -0,0 +1,278 @@
+.. _radial-integral:
+
+Concepts behind the radial integral
```
Luthaf, line 3:
```suggestion
SOAP and LODE radial integrals
```

docs/src/explanations/radial-integral.rst
```diff
@@ -0,0 +1,278 @@
...
+       I_{nl}(r_{ij}) = \frac{1}{(\pi \sigma^2)^{3/4}} \cdot
+       \pi^{\frac{3}{2}}\frac{\Gamma\left(n_\mathrm{eff}\right)}{\Gamma\left(l_\mathrm{eff}\right)}\frac{(ar_{ij})^l}{(a+b)^{n_\mathrm{eff}}}M\left(n_\mathrm{eff},l_\mathrm{eff},\frac{a^2r_{ij}^2}{a^2+b^2}\right),
+   \end{aligned}
+
+where :math:`M(a,b,z)` is the confluent hypergeometric function (hyp1f1).
+
+.. _k-space-radial-integral-1:
+
+K-space Radial Integral
+-----------------------
```
Luthaf, line 278: Is this missing?

docs/src/explanations/radial-integral.rst
```diff
@@ -0,0 +1,278 @@
+.. _radial-integral:
```
Luthaf, line 1: I like having this page around, but it does not flow very well yet (in  part because it assumes knowledge of spherical expansion & friends, which should be explained in a different section). We can still merge it now and improve it later!

python/rascaline/rascaline/utils/atomic_density.py
```diff
@@ -0,0 +1,166 @@
...
+    :param potential_exponent: Potential exponent of the decorated atom density.
+        Currently only implemented for potential_exponent < 10. Some exponents can be
+        connected to SOAP or physics-based quantities: p=0 uses Gaussian densities as in
+        SOAP, p=1 uses 1/r Coulomb like densities, p=6 uses 1/r^6 dispersion like
+        densities.
+    """
+
+    def __init__(self, atomic_gaussian_width: float, potential_exponent: int):
+        if not HAS_SCIPY:
+            raise ValueError("LodeDensity requires scipy!")
```
Luthaf, line 141:
```suggestion
            raise ValueError("LodeDensity requires scipy to be installed")
```

python/rascaline/rascaline/utils/atomic_density.py
```diff
@@ -0,0 +1,166 @@
...
+
+        self.potential_exponent = potential_exponent
+        self.atomic_gaussian_width = atomic_gaussian_width
+
+    def _f_sr(self, a, x):
```
Luthaf, line 146:
```suggestion
    def _short_range(self, a, x):
```

python/rascaline/rascaline/utils/atomic_density.py
```diff
@@ -0,0 +1,166 @@
...
+        return 1 / a - x / (a + 1) + x**2 / (2 * (a + 2))
+
+    def _f_lr(self, a, x):
```
Luthaf, line 149:
```suggestion
    def _long_range(self, a, x):
```

python/rascaline/rascaline/utils/atomic_density.py
```diff
@@ -0,0 +1,166 @@
...
+        if self.potential_exponent == 0:
+            return GaussianDensity.compute(self, positions=positions)
+        else:
+            atomic_gaussian_width_sq = self.atomic_gaussian_width**2
+            a = self.potential_exponent / 2
+            x = positions**2 / (2 * atomic_gaussian_width_sq)
+
+            prefac = 1 / gamma(a) / (2 * atomic_gaussian_width_sq) ** a
+
+            return prefac * np.where(
```
Luthaf, line 162: Could you explain why you use a different implementation for short and long range here? And how the 1e-5 was picked?

python/rascaline/rascaline/utils/atomic_density.py
```diff
@@ -0,0 +1,166 @@
...
+    HAS_SCIPY = True
+except ImportError:
+    HAS_SCIPY = False
+
+
+class AtomicDensityBase(ABC):
+    """Base class implemententing atomic densities."""
+
+    @abstractmethod
+    def compute(self, positions: Union[float, np.ndarray]) -> Union[float, np.ndarray]:
```
Luthaf, line 52: We could use `@overload` to do something like

```
def compute(self, positions: float) -> float:

def compute(self, positions: np.ndarray) -> np.ndarray:
```

python/rascaline/rascaline/utils/radial_basis.py
```diff
@@ -0,0 +1,325 @@
+r"""
+.. _python-radial-basis:
+
+Radial Basis
+============
+
+Radial basis functions :math:`R_{nl}(\boldsymbol{r})` are besides :ref:`atomic densities
+<python-atomic-density>` :math:`\rho_i` the central ingridents to compute spherical
```
Luthaf, line 8:
```suggestion
<python-atomic-density>` :math:`\rho_i` the central ingredents to compute spherical
```

python/rascaline/tests/utils/splines.py
```diff
@@ -102,3 +111,261 @@ def test_splines_numerical_derivative_error():
...
+                * kk**ell
+                * hyp1f1(i1, i2, -0.5 * (kk * sigma[n]) ** 2)
+            )
+
+            coeffs_num[n, ell] = spliner._radial_integral(n, ell, kk)
+
+    assert_allclose(coeffs_num, coeffs_exact)
+
+
+# def test_rspace_radial_integral():
```
Luthaf, line 170: why is this commented out?

python/rascaline/tests/utils/splines.py
```diff
@@ -102,3 +111,261 @@ def test_splines_numerical_derivative_error():
...
+    k_cutoff = 1.2 * np.pi / atomic_gaussian_width
+
+    basis = GtoBasis(cutoff=cutoff, max_radial=max_radial)
+
+    spliner = KSpaceSpliner(
+        max_radial=max_radial,
+        max_angular=max_angular,
+        k_cutoff=k_cutoff,
+        basis=basis,
+        density=DeltaDensity(),  # density does not enter in a Kspace radial integral
```
Luthaf, line 132: if the atomic density does not enter the K-space RI, should we remove the parameter from the constructor?

python/rascaline/tests/utils/splines.py
```diff
@@ -102,3 +111,261 @@ def test_splines_numerical_derivative_error():
...
+        "cutoff": cutoff,
+        "max_radial": max_radial,
+        "max_angular": max_angular,
+        "center_atom_weight": 1.0,
+        "atomic_gaussian_width": atomic_gaussian_width,
+        "cutoff_function": {"Step": {}},
+    }
+
+    analytic = SphericalExpansion(
+        radial_basis={"Gto": {}}, **hypers_spherical_expansion
```
Luthaf, line 279: you can also use accuracy 1e-4 here

python/rascaline/rascaline/utils/splines.py
```diff
@@ -401,3 +479,380 @@ def _radial_integral_derivative(
             return super()._radial_integral_derivative(n, ell, positions)
         else:
             return self.radial_integral_derivative_funcion(n, ell, positions)
+
+
+class RealSpaceSpliner(RadialIntegralSplinerBase):
```
Luthaf, line 484: Should these be called `SoapSpliner` and `LodeSpliner` instead?

Luthaf (commented):
Looks mostly good. I'll try to reflow the docs a bit, but that's easier to do as a commit instead of github suggestions.

python/rascaline/rascaline/utils/splines.py
```diff
@@ -1,18 +1,43 @@
...
+.. autoclass:: rascaline.utils.KSpaceSpliner
+    :members:
+    :show-inheritance:
+
+Note that :class:`RealSpaceSpliner` and :class:`KSpaceSpliner` require `scipy`_ to
+be installed in order to perform the numercial integrals.
+
+Besides the two classes you can also explicitly provide functions for the radial
+integral and its derivive and passing them to
+
```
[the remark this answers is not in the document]
PicoCentauri: Yes, I thought the same but was lazy to change it. Thanks!

docs/src/explanations/radial-integral.rst
PicoCentauri: Yes, we have to write it 🤣 pinging @kvhuguenin here because heis the master for this. We can remove the section for now.

python/rascaline/rascaline/utils/atomic_density.py
PicoCentauri: But then we have to provide both functions in all subclasses no?

docs/src/explanations/radial-integral.rst
Luthaf: Might be worth to move it to some "advanced documentation" section

python/rascaline/rascaline/utils/atomic_density.py
PicoCentauri: Because if you do not do this you get `0/0` terms. I tried to explain this in docstring.

python/rascaline/tests/utils/splines.py
PicoCentauri: But is does for the center contribution which we are not testing here.

python/rascaline/tests/utils/splines.py
PicoCentauri: Becuase, I still need to fix the analytical expression. 🫣

python/rascaline/tests/utils/splines.py
PicoCentauri: ohh yes, this might be better!

docs/src/explanations/radial-integral.rst
PicoCentauri: Yes, fine for me.

PicoCentauri: I moved it to the devdocs. I think there it make more sense in its current state. I also added a sentence:


> Note that this page assumes knowledge of spherical expansion & friends and currently serves as
> a reference page for the developers to support the implementation.

PicoCentauri:
Thanks again for the feedback @Luthaf. If you want to give the docs another round of improvements I am happy if you provide a commit. Otherwise we can also proceed in another PR. This is already quite big...

Luthaf:
CI error is `pulling from host quay.io failed with status code [manifests latest]: 500 Internal Server Error`, I'll try to re-run later.

PicoCentauri (commented):
Thanks for the really nice improvements @Luthaf !!!

I am happy with the PR and hope that this can be helpful and gives us more flexibility of the basis and density combinations.

PR #185, opened 2023-05-23
Added cg_utils to Rascaline

ecignoni:
----
:books: Documentation preview :books:: https://rascaline--185.org.readthedocs.build/en/185/

PR #237, opened 2023-09-28
Redesigned metatensor-baseed Clebsch Gordan iterations

### LINKED ISSUE #178: Clebsch-Gordan utils
We need to add the Clebsch-Gordan (CG) utils somewhere, so we decided to add it to rascaline.

### Implementation
There are severe problems with all existing CG utils which need to be addressed however. @Luthaf had some thoughts about it.

### Mathematical documentation
As far as I know the most recent version of the recursive relationship of ACDC features is in the supplementary of the "Optimal radial basis for density-based atomic representations" paper.
Link: https://pubs.aip.org/jcp/article-supplement/1013253/zip/104106_1_supplements/
See Section VII. A SIMPLIFIED DERIVATION OF THE NICE ITERATION

@ecignoni is working on simplified math paper that explains the key points important to derive the equivariant representations that might be more useful for a doc page in rascaline, since the before-mentioned supplementary is quite technical. Will give an update in two weeks

jwa7:
Closes #178, and invalidates #185 . New API design for performing metatensor-based Clebsch Gordan iterations.

Still left to do:

- [ ] Mathematical unit tests (handled by @agoscinski and @DivyaSuman14)
- [ ] Unit tests on the pre-computation of metadata (@jwa7)
- [ ] Thorough documentation


agoscinski:
Should I add the doc from PR https://github.com/Luthaf/rascaline/pull/183? That seemed to me as a good starting point

jwa7:
> Should I add the doc from PR #183? That seemed to me as a good starting point

Yeah sounds like a good idea to inherit the .rst file from there. Is there a way to pipe the relevant commit(s) from that branch into this PR so credit can be properly assigned?

Luthaf:
You can use `git cherry-pick` to re-apply individul commits on top of a new branch. Alternatively, you can create a new commit and set a different author than yourself, or just use `Co-Authored-By` in the commit message.

jwa7:
Update: @curiosity54 and I have fixed the norm preservation unit test!  I think we are in a position to discuss finalizing the API and then moving onto documentation. Thoughts @luthaf?

jwa7:
And to be clear on what happened: I have enabled a bool flag `sort_l_list` (open to suggestions on the name) that either turns on/off the computation of redundant blocks with a CG iteration - i.e. ones in which the sorted l lists are the same.

When this flag is off, the norm preservation tests pass as expected. When switxhed on, thus saving computational cost, the test failed too as expected.

The alternative was a tricky handling of the l list with the normalization. This might be more tricky than first thought, as also depends on k-combinations, and as such was not something Jigyasa and I could figure out just now.

As this normalization (in the case of taking computational shortcuts) perhaps isn't important for ML purposes, are we ok to let users use turn on this flag at their own risk - i.e. accept the slight information loss at reduced computational cost?

Luthaf (commented):
small comments, I'll do a full review later

jwa7:
@Luthaf In order to make the CI happy I had to add some dependencies in the min-deps env - for CG tests we need `ase`, `metatensor-operations` and `sympy`. Of course not ideal as defeats the purpose of being minimal dependencies - what's the best way to get around this? Is there a way to only run CG tests at a higher level of dependencies?

Luthaf:
>  Is there a way to only run CG tests at a higher level of dependencies?

yes! https://docs.pytest.org/en/7.1.x/how-to/skipping.html#skipping-on-a-missing-import-dependency

jwa7:
Thanks for the review @Luthaf ! I've made the changes re: your suggestions :)

Luthaf (approved):

PR #248, opened 2023-10-18
Add a (-1)^l factor to the power spectrum invariants

Luthaf:
This make the power spectrum produce the same output as doing a full Clebsch-Gordan product and keeping only the lambda=0 term

----
:books: Documentation preview :books:: https://rascaline--248.org.readthedocs.build/en/248/

### DIFF
--- a/python/rascaline/rascaline/utils/power_spectrum/calculator.py
+++ b/python/rascaline/rascaline/utils/power_spectrum/calculator.py
@@ -217,7 +217,10 @@ class PowerSpectrum:
             ell = key[0]
             species_center = key[1]
 
-            factor = 1 / sqrt(2 * ell + 1)
+            # For consistency with a full Clebsch-Gordan product we need to add
+            # a `-1^l / sqrt(2 l + 1)` factor to the power spectrum invariants
+            factor = (-1) ** ell / sqrt(2 * ell + 1)
+
             # Find that block indices that have the same spherical_harmonics_l and
             # species_center
             selection = Labels(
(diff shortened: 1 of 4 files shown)


jwa7 (approved):
Performed the consistency check of the invariant of lambda-SOAP vs Python PowerSpectrum implementation and it passes, as if the CI looks happy it looks good to me!

python/rascaline/rascaline/utils/power_spectrum/calculator.py
```diff
@@ -217,7 +217,8 @@ def compute(
             ell = key[0]
             species_center = key[1]
 
-            factor = 1 / sqrt(2 * ell + 1)
+            factor = (-1) ** ell / sqrt(2 * ell + 1)
```
PicoCentauri, line 220: I would add the comment from the rust code here as well:

```suggestion
             # For consistency with a full Clebsch-Gordan product we need to add
             # a `-1^l / sqrt(2 l + 1)` factor to the power spectrum invariants
            factor = (-1) ** ell / sqrt(2 * ell + 1)
```

rascaline/src/calculators/soap/power_spectrum.rs
```diff
@@ -375,6 +386,8 @@ impl SoapPowerSpectrum {
 struct SpxPropertiesToCombine<'a> {
     /// value of l
     spherical_harmonics_l: usize,
+    /// normalization factor (-1)^l * sqrt(2 l + 1)
```
PicoCentauri, line 389: use the power of LateX
```suggestion
    /// normalization factor :math:`-1^l  \sqrt{2 l + 1}`
```

scripts/clean-python.sh
```diff
@@ -10,6 +10,7 @@ cd "$ROOT_DIR"
 
 rm -rf dist
```
PicoCentauri, line 11: Ahh there is file to clean the Python build leftovers. Very useful!

PicoCentauri (changes requested):
Some minor docs comment, otherwise I am happy. I like when we break the code.

PicoCentauri (approved):
Ship it! 🚢

PR #253, opened 2023-10-30
Use rascaline calculators inside metatensor's atomistic models

Luthaf:
This PR does two things:

- replace `rascaline.torch.System` with `metatensor.torch.atomistic.System`
- implement `requested_neighbors_lists` to tell the simulation engine which neighbors lists we need for rascaline

This builds on top of https://github.com/lab-cosmo/metatensor/pull/405

----
:books: Documentation preview :books:: https://rascaline--253.org.readthedocs.build/en/253/


docs/src/references/api/torch/systems.rst
```diff
@@ -1,7 +1,4 @@
 System
 ======
 
-.. autoclass:: rascaline.torch.System
```
PicoCentauri, line 4: Maybe add a comment that we are using the metatensor systems class and we provide some conversion functions.

PicoCentauri (commented):
A quick comment. I do a deeper review asap.

PicoCentauri (approved):
Besides my comment to the docs it looks very good.

docs/src/references/api/torch/systems.rst
Luthaf: done!

PicoCentauri (approved):

PR #256, opened 2023-11-03
Fix access to rascaline-torch cmake_prefix_path from Python

Luthaf:
It was lost when adding the Python Power Spectrum, since we had both `utils.py` and `utils/__init__.py`

----
:books: Documentation preview :books:: https://rascaline--256.org.readthedocs.build/en/256/


PicoCentauri (commented):
Maybe can we add a test that this exist to prevent this in the future?

python/rascaline-torch/tests/utils/cmake_prefix.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+import rascaline.torch
+
+
+def test_cmake_prefix():
+    assert os.path.exists(rascaline.torch.utils.cmake_prefix_path)
```
Luthaf, line 7: @PicoCentauri here is the test checking that we can get the python variable, and the it points to something reasonable

PicoCentauri, line 7: 😍

PicoCentauri (approved):

PR #260, opened 2023-11-09
Fixed double-counting of periopdic image pairs

Luthaf:
This was found by @liam-o-marsh in #240, and I cherry-picked it to merge the fix quicker. 

Ping @rosecers who wants to use this!

----
:books: Documentation preview :books:: https://rascaline--260.org.readthedocs.build/en/260/

### DIFF
--- a/python/rascaline/rascaline/systems/ase.py
+++ b/python/rascaline/rascaline/systems/ase.py
@@ -91,10 +91,23 @@ class AseSystem(SystemBase):
 
         nl_result = neighborlist.neighbor_list("ijdDS", self._atoms, cutoff)
         for i, j, d, D, S in zip(*nl_result):
+            # we want a half neighbor list, so drop all duplicated neighbors
             if j < i:
-                # we want a half neighbor list, so drop all duplicated
-                # neighbors
                 continue
+            elif i == j:
+                if S[0] == 0 and S[1] == 0 and S[2] == 0:
+                    # only create pairs with the same atom twice if the pair spans more
+                    # than one unit cell
+                    continue
+                elif S[0] + S[1] + S[2] < 0 or (
+                    (S[0] + S[1] + S[2] == 0) and (S[2] < 0 or (S[2] == 0 and S[1] < 0))
+                ):
+                    # When creating pairs between an atom and one of its periodic
+                    # images, the code generate multiple redundant pairs (e.g. with
+                    # shifts 0 1 1 and 0 -1 -1); and we want to only keep one of these.
+                    # We keep the pair in the positive half plane of shifts.
+                    continue
+
             self._pairs.append((i, j, d, D, S))
 
         self._pairs_by_center = []
--- a/python/rascaline/tests/systems/ase.py
+++ b/python/rascaline/tests/systems/ase.py
@@ -1,6 +1,7 @@
 import numpy as np
 import pytest
 
+from rascaline import SphericalExpansion
 from rascaline.systems import AseSystem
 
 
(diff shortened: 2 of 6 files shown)


rosecers:
Much appreciated on my end! Code looks great to this noob -- let me know if you want a formal review from me.

Luthaf:
I think the NeighborsList calculator is now wrong with these changes, around here: https://github.com/Luthaf/rascaline/blob/2851d47ab3471057f3794ac2b48b0417add2f748/rascaline/src/calculators/neighbor_list.rs#L568-L571

The condition should account for the cell shift as well.

python/rascaline/rascaline/systems/ase.py
```diff
@@ -91,10 +91,23 @@ def compute_neighbors(self, cutoff):
...
         nl_result = neighborlist.neighbor_list("ijdDS", self._atoms, cutoff)
         for i, j, d, D, S in zip(*nl_result):
+            # we want a half neighbor list, so drop all duplicated neighbors
             if j < i:
-                # we want a half neighbor list, so drop all duplicated
-                # neighbors
                 continue
+            elif i == j:
+                if S[0] == 0 and S[1] == 0 and S[2] == 0:
+                    # only create pair with the same atom twice if the pair spans more
```
PicoCentauri, line 99:
```suggestion
                    # only create pairs with the same atom twice if the pair spans more
```

python/rascaline/tests/systems/ase.py
```diff
@@ -145,3 +146,54 @@ def test_no_pbc_cell():
...
+            (2.19266, 2.11524, 3.86858),
+            (2.52936, 4.62777, 3.87771),
+            (2.01817, 0.85408, 3.87087),
+        ],
+        cell=[[3.57941, 0, 0], [0.682558, 5.02733, 0], [0.285565, 0.454525, 7.74858]],
+        pbc=True,
+    )
+
+    calculator = SphericalExpansion(
+        cutoff=9,
```
PicoCentauri, line 168:
```suggestion
        cutoff=9,  # choose a cutoff that is larger then the cell to test for "ghost" atoms
```

python/rascaline/tests/systems/ase.py
```diff
@@ -145,3 +146,54 @@ def test_no_pbc_cell():
     )
     with pytest.warns(Warning, match=message):
         AseSystem(atoms)
+
+
+def test_same_spherical_expansion():
+    system = ase.Atoms(
```
PicoCentauri, line 152: I have the feeling that this cell is this to small to check for more then one unit cell?

rascaline/src/systems/neighbors.rs
```diff
@@ -250,10 +250,28 @@ impl CellList {
...
+                                        continue;
+                                    }
+
+                                    if (shift[0] + shift[1] + shift[2] < 0) ||
+                                        (
+                                            (shift[0] + shift[1] + shift[2] == 0) &&
+                                            (shift[2] < 0 || (shift[2] == 0 && shift[1] < 0))
+                                        )
+                                    {
+                                        // When creating pairs between an atom
```
PicoCentauri, line 266: I understand what is going on here but maybe a picture would help. But I agree and accept, if this is too much work.

PicoCentauri (commented):
Good to have this fixed.

python/rascaline/tests/systems/ase.py
Luthaf: It's 3.6 in one direction, you should get pairs with a shift of 2 with a cutoff of 9

rascaline/src/systems/neighbors.rs
liam-o-marsh: does this work?

```
──────┼─────┼─────┼─────┼─────┼
     ╭╯    ╭╯    ╭╯    ╭╯    ╭╯
     │     │     │   × │     │
     │     │     │  ╱  │     │
    ╭╯    ╭╯    ╭╯ ╱  <- first pair
────┼─────┼─────┼─────┼─────┼
   ╭╯    ╭╯   ╱╭╯    ╭╯    ╭╯
   │     │   × │     │     │
   │     │  ╱  │     │     │
  ╭╯    ╭╯ ╱  <- second pair, redundant
──┼─────┼─────┼─────┼─────┼
 ╭╯    ╭╯    ╭╯    ╭╯    ╭╯
 │    ╱│     │     │     │
 │   × │     │     │     │
╭╯    ╭╯    ╭╯    ╭╯    ╭╯
┼─────┼─────┼─────┼─────┼
```

Luthaf: We looked at the code very carefully again, and I think we both understand it now! I might do a diagram of the 2D version of it, but trying to represent the 3D version in ASCII is a bit clunky

liam-o-marsh: I think I understand what the hard part is in the understanding. my reasoning is:
- for two same-atom pairs characterised by (h,j,k) and (-h,-j,-k), one of them has to be ignored. how do you choose?
- first try to divide space along the h+j+k=0 plane.
- if it doesn't work, divide by the k=0 plane
- if it still doesn't work, divide by the j=0 plane.
- if the two pairs cannot be separated by any of those planes, then they are both at (h=0,j=0,k=0), which isn't possible

Luthaf: Yes, that was the issue. I split the condition in two and added a small illustration of the 2D case

PicoCentauri (approved):

PR #263, opened 2023-12-07
added pyscf system

liam-o-marsh:
----
:books: Documentation preview :books:: https://rascaline--263.org.readthedocs.build/en/263/

### DIFF
--- a/python/rascaline/rascaline/systems/__init__.py
+++ b/python/rascaline/rascaline/systems/__init__.py
@@ -35,13 +35,14 @@ if HAVE_CHEMFILES:
 
 if HAVE_PYSCF:
     IntoSystem.__doc__ += """
-    - `pyscf.gto.mole.Mole`_ or `pyscf.pbc.gto.cell.Cell`_: pyscf' Frame type. There is no associated neighbor
-      list implementation, the system will only be usable with
-      ``use_native_system=True``
+    - `pyscf.gto.mole.Mole`_ or `pyscf.pbc.gto.cell.Cell`_: pyscf' Frame type.
+       There is no associated neighbor list implementation, the system will only
+       be usable with ``use_native_system=True``
 
     .. _pyscf.gto.mole.Mole: https://pyscf.org/user/gto.html
     .. _pyscf.pbc.gto.cell.Cell: https://pyscf.org/user/pbc/gto.html
     """
+
 
 def wrap_system(system: IntoSystem) -> SystemBase:
     """Wrap different systems implementation into the right class.
@@ -63,7 +64,7 @@ def wrap_system(system: IntoSystem) -> SystemBase:
 
     if HAVE_ASE and AseSystem.can_wrap(system):
         return AseSystem(system)
-        
+
     if HAVE_PYSCF and PyscfSystem.can_wrap(system):
         return PyscfSystem(system)
 
--- a/python/rascaline/rascaline/systems/__init__.py
+++ b/python/rascaline/rascaline/systems/__init__.py
@@ -1,6 +1,7 @@
 from .ase import HAVE_ASE, AseSystem
 from .base import SystemBase
 from .chemfiles import HAVE_CHEMFILES, ChemfilesSystem
+from .pyscf import HAVE_PYSCF, PyscfSystem
 
 
 class IntoSystem:
@@ -32,6 +33,15 @@ if HAVE_CHEMFILES:
     .. _chemfiles.Frame: http://chemfiles.org/chemfiles.py/latest/reference/frame.html
     """
 
+if HAVE_PYSCF:
+    IntoSystem.__doc__ += """
+    - `pyscf.gto.mole.Mole`_ or `pyscf.pbc.gto.cell.Cell`_: pyscf' Frame type. There is no associated neighbor
+      list implementation, the system will only be usable with
+      ``use_native_system=True``
+
+    .. _pyscf.gto.mole.Mole: https://pyscf.org/user/gto.html
+    .. _pyscf.pbc.gto.cell.Cell: https://pyscf.org/user/pbc/gto.html
+    """
 
 def wrap_system(system: IntoSystem) -> SystemBase:
     """Wrap different systems implementation into the right class.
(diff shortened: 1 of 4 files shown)


python/rascaline/rascaline/systems/__init__.py
```diff
@@ -32,6 +33,15 @@ def __init__(self):
     .. _chemfiles.Frame: http://chemfiles.org/chemfiles.py/latest/reference/frame.html
     """
 
+if HAVE_PYSCF:
```
Luthaf, line 36: Could you add pyscf to the `docs` env in tox.ini, to make sure this is included in the documentation?

python/rascaline/rascaline/systems/pyscf.py
```diff
@@ -0,0 +1,102 @@
...
+        self.is_periodic = isinstance(self._frame, pyscf.pbc.gto.cell.Cell)
+
+    def size(self):
+        return self._frame.natm
+
+    def species(self):
+        return self._species
+
+    def positions(self):
+        return self._frame.atom_coords()
```
Luthaf, line 81: just to check, is this using cartesian or fractional coordinates?

python/rascaline/rascaline/systems/pyscf.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def cell(self):
+        if self.is_periodic:
+            return self._frame.a
```
Luthaf, line 85: just to check, the cell is stored in row or column-major order? I.e. is the first cell vector `cell[0, :]` or `cell[:, 0]`?

Luthaf (commented):
Looks good overall, just a handful of questions!

python/rascaline/rascaline/systems/__init__.py
Luthaf: It would be nice to also add a test (and add pyscf in the `all-deps` env in tox) checking that you get the right data out of both PySCF classes, for both molecules and periodic systems.

python/rascaline/rascaline/systems/pyscf.py
liam-o-marsh: I already checked that this matches, but I can't remember which is the shared order

liam-o-marsh: from the pyscf docs: "Each row of the 3-by-3 matrix of a represents a lattice vector in Cartesian coordinate, with the same unit as the input atom parameter." (same unit because pyscf stores everything in bohrs internally, but by default the coordinates you specify are in angströms)

liam-o-marsh:
(the tests are currently a WIP, since I didn't get to run them yet)

python/rascaline/rascaline/systems/pyscf.py
DivyaSuman14: I think atom_coords() return cartesian coordinates (in units of Bohr)

Luthaf:
I though PySCF was a pure Python project, this complicates things a lot! From looking through the issues on PySCF github, it seems that adding a BLAS implementation will not solve the issue here (the code later tries to call `make`, which does not exists on Windows). 

A better solution might be to only add the dependency on *nix platforms, replacing `pyscf` with `pyscf;platform_system!="Windows"` (i.e. using [environment-markers](https://peps.python.org/pep-0508/#environment-markers)) in the `deps`.

Luthaf:
Why not! I looked for something like it but could not find it. I would name the custom platforms something closer to this example: https://tox.wiki/en/4.11.4/faq.html#platform-specification (or using `linux_: linux`, `windows_: win32,…`).

I would also not touch `mops` yet, it has a couple of other issues in addition to windows support and is still in alpha! But its good to know we could do this!

liam-o-marsh:
in the end, what I was suggesting didn't work, it would have needed to define [tesenv:all-deps-lin],[tesenv:all-deps-mac] and [tesenv:all-deps-win] separately
however, I did manage to mix pip environment markers with url-based dependencies.

liam-o-marsh:
don't approve the workflows just yet, I have to debug the CI, and I have another PR for that. I'll tell you when the debugging is done

liam-o-marsh:
alright, we are good to go

liam-o-marsh:
...and two CI failures that are not the code's fault ("failed to submit to codecov" and "readthedocs build timed out")

Luthaf:
Yes, I'm restarting them to get them to work. Codecov in particular has been painful to deal with for while.

python/rascaline/rascaline/systems/pyscf.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def cell(self):
+        if self.is_periodic:
+            cell = self._frame.a
+            if self._frame.unit[0].lower() == "a":
+                # assume angströms, we are good to go
+                return cell
+            else:
+                # assume bohrs, correct this
+                return pyscf.data.nist.BOHS * np.asarray(cell)
```
Luthaf, line 105: Should there be similar logic for positions?

python/rascaline/tests/systems/pyscf.py
```diff
@@ -0,0 +1,47 @@
...
+    )
+    print(system.positions(), positions)
+    assert np.allclose(system.positions(), positions, rtol=1e-14)
+    assert np.all(system.cell() == [[0, 0, 0], [0, 0, 0], [0, 0, 0]])
+
+
+def test_pbc_data():
+    import pyscf.pbc
+
+    atoms = pyscf.pbc.gto.Cell(
```
Luthaf, line 38: could you add a test with explicit units if that's something PySCF supports?

python/rascaline/rascaline/systems/pyscf.py
liam-o-marsh: no, the positions are obtained from pyscf's internal representation (directly in Bohrs) while the cell is as-given-to-pyscf, so its unit can be either Bohrs or Angströms.

python/rascaline/tests/systems/pyscf.py
liam-o-marsh: done

python/rascaline/rascaline/systems/pyscf.py
Luthaf: that's a weird API from PySCF, but ok ^_^

liam-o-marsh: well, to be fair, I am not using "the proper API" to fetch the boundary information, but I don't think there *is* a proper API for this specific information.
...though there is mol._h which is always in Bohrs, and is the transposed version of what we want.

Luthaf: ok, make sense!

Luthaf:
Thanks for doing this!

liam-o-marsh:
you're welcome!

PR #273, opened 2024-01-24
Add `species` option to `PowerSpectrum` class

PicoCentauri:
Also I moved the docstring to the class itself and also fixed a typo in another doc file.

----
📚 Documentation preview 📚: https://rascaline--273.org.readthedocs.build/en/273/


rascaline/Cargo.toml
```diff
@@ -58,6 +58,7 @@ approx = "0.5"
 
 [dev-dependencies]
 criterion = "0.5"
+half = "2.2.1"
```
Luthaf, line 61: maybe merge this with the block below about `clap`?
```suggestion
half = "=2.2.1"
```

PicoCentauri: Yes, that was the plan I don't why this is gone.

python/rascaline-torch/rascaline/torch/utils/power_spectrum.py
```diff
@@ -58,6 +58,9 @@ class PowerSpectrum(torch.nn.Module, PowerSpectrumBase):
 
     :param calculator_1: first calculator
     :param calculator_1: second calculator
+    :param species: List of `species_neighbor` to fill all blocks with. This option
+        might be useful when joining along the ``sample`` direction after computation.
+        If :py:obj:`None` blocks are filled with `species_neighbor` from all blocks.
```
Luthaf, line 63: I feel like this is a bit clearer, what do you think?

```suggestion
    :param species: List of `species_neighbor` to use in the properties of the output. 
        This option might be useful when running the calculation on subset of the whole dataset and trying to join along the ``sample`` dimension after the calculation.
        If :py:obj:`None` blocks are filled with `species_neighbor` found in the systems.
```

PicoCentauri: Yes I agree. Thanks!

PR #275, opened 2024-02-02
fix broken metatensor documentation links

PicoCentauri:
----
📚 Documentation preview 📚: https://rascaline--275.org.readthedocs.build/en/275/

### DIFF
--- a/docs/src/explanations/concepts.rst
+++ b/docs/src/explanations/concepts.rst
@@ -67,7 +67,7 @@ After using a calculator on one or multiple systems, users will get the
 numerical representation of their atomic systems in a ``descriptor`` object.
 Rascaline uses `metatensor`_ ``TensorMap`` type when returning descriptors.
 
-.. _metatensor: https://lab-cosmo.github.io/metatensor/latest/
+.. _metatensor: https://lab-cosmo.github.io/metatensor/
 
 A ``TensorMap`` can be seen as a dictionary mapping some keys to a set of data
 blocks. Each block contains both data (and gradients) arrays — i.e.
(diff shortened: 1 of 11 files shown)


Luthaf:
I think I would rather have `https://lab-cosmo.github.io/metatensor/` redirect to `https://lab-cosmo.github.io/metatensor/latest/` on metatensor side. This should be doable with a minimal index.html which directly redirects (I though we already did something like this)

PicoCentauri:
A redirect is also fine for me.

docs/src/get-started/installation.rst
```diff
@@ -175,5 +175,5 @@ Other useful configuration options are:
 |                                        | install metatensor-torch?                     |                |
 +----------------------------------------+-----------------------------------------------+----------------+
 
-.. _C++ interface of metatensor: https://lab-cosmo.github.io/metatensor/latest/get-started/installation.html#installing-the-c-and-c-library
-.. _TorchScript interface of metatensor: https://lab-cosmo.github.io/metatensor/latest/get-started/installation.html#for-usage-from-c
+.. _C++ interface of metatensor: https://lab-cosmo.github.io/metatensor/get-started/installation.html#installing-the-c-and-c-library
+.. _TorchScript interface of metatensor: https://lab-cosmo.github.io/metatensor/get-started/installation.html#for-usage-from-c
```
Luthaf, line 179: I think these still need to contain the `latest`. The redirect only happen on direct access to the `metatensor/` url.

PicoCentauri: Yes, right.

Luthaf (approved):

PR #278, opened 2024-02-07
Use the input dtype and device for the output in rascaline-torch

Luthaf:
All calculations still happen on CPU with 64-bit floats, but the data is automatically converted on input and output as required.

Should help with https://github.com/lab-cosmo/metatensor-models/pull/58

----
📚 Documentation preview 📚: https://rascaline--278.org.readthedocs.build/en/278/


frostedoyster (approved):
Tested and it works. Thanks a lot!

PR #292, opened 2024-03-25
Parallelize compilation with cmake

frostedoyster:
We use cmake's `--parallel` flag to speed up compilation for Python builds

----
📚 Documentation preview 📚: https://rascaline--292.org.readthedocs.build/en/292/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -112,11 +112,11 @@ class cmake_ext(build_ext):
             check=True,
         )
         subprocess.run(
-            ["cmake", "--build", build_dir],
+            ["cmake", "--build", build_dir, "--parallel"],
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -112,11 +112,11 @@ def run(self):
             check=True,
         )
         subprocess.run(
-            ["cmake", "--build", build_dir],
+            ["cmake", "--build", build_dir, "--parallel"],
             check=True,
         )
```
Luthaf, line 117: Let's remove this whole thing and let the one below take care of the build and the install.

frostedoyster: Okk

PR #293, opened 2024-03-28
Parallelize LODE calculation over samples

Luthaf:
This is only used when we have fewer systems than available threads.

For many system, this is between 5% slower and 20% faster than the previous code (likely due to the reorganization of the loop over samples). For a single system, this is up to 6x faster with 10 threads for the molecular_crystal benchmark file.

Requested by @andreagrisafi for use in SALTED!

----
📚 Documentation preview 📚: https://rascaline--293.org.readthedocs.build/en/293/

### DIFF
--- a/rascaline/src/calculators/lode/spherical_expansion.rs
+++ b/rascaline/src/calculators/lode/spherical_expansion.rs
@@ -85,6 +85,8 @@ pub struct LodeSphericalExpansion {
     /// The vector contains different l values, and the Array is indexed by
     /// `m, n, k`.
     k_vector_to_m_n: ThreadLocal<RefCell<Vec<Array3<f64>>>>,
+    /// Cached allocation for everything that only depends on the k vector
+    k_dependent_values: ThreadLocal<RefCell<Array1<f64>>>,
 }
 
 /// Compute the trigonometric functions for LODE coefficients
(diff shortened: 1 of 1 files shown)


rascaline/src/calculators/lode/spherical_expansion.rs
```diff
@@ -578,8 +591,25 @@ impl CalculatorBase for LodeSphericalExpansion {
...
+        // Trying to use both at the same time would require additional
+        // synchronisation to ensure the right threads access per-system
+        // pre-computed data (i.e. anything that depends on k-vectors).
+        let (parallel_systems, parallel_samples) = if systems.len() > rayon::current_num_threads() {
+            (true, false)
+        } else {
+            (false, true)
+        };
+        let n_systems = systems.len();
+
```
PicoCentauri, line 609: I think this is something we should out at a more general location in the code, so that all calculators can use this in the future. Maybe in `CalculatorBase`?

PicoCentauri (commented):
Looks very good. The code also didn't became much more complicated. I have a minor comment but we can also do this in a later commit.

rascaline/src/calculators/lode/spherical_expansion.rs
Luthaf: I don't agree: the optimal strategy might depend on the algorithms, so this should be decided close to where the algorithms are executed.

PR #298, opened 2024-04-12
Add radial integral for custom densities

PicoCentauri:
This PR adds the ability to compute the radial integral for custom densities that have a spherical symmetry. Spherical symmetry should apply to all densities and this PR will be especially useful to compute a real space version of a spherical expansion with a LODE density and any basis.

The function computes the double integral numerically using scipy's `dblquad` function and I expect that this will very slow. Maybe we should add a warning?

# TODO

- [x] Add required `compute_derivative` for `GaussianDensity` (only needed for tests) and `LodeDensity`
- [x] Tests: My idea is to test the custom radial integral with a `GaussianDensity` density against the analytical version that we already have implemented.

----
📚 Documentation preview 📚: https://rascaline--298.org.readthedocs.build/en/298/


python/rascaline/tests/utils/atomic_density.py
```diff
@@ -0,0 +1,25 @@
...
+    ],
+)
+def test_derivative(density):
+    # Exclude very small values due instabilitues of LodeDensity
+    positions = np.linspace(1e-7, 5, num=1000)
+    dens = density.compute(positions)
+    grad_ana = density.compute_derivative(positions)
+    grad_numerical = np.gradient(dens, positions)
+
+    assert_allclose(grad_numerical, grad_ana, atol=1e-3)
```
Luthaf, line 25: this tolerance is pretty high =/. I tend to go for atol~1e-9, rtol~1e-6 ideally. Here if all the values are below 1e-3, the test will do nothing.

PicoCentauri: Sure, I can increase it!

python/rascaline/tests/utils/atomic_density.py
```diff
@@ -0,0 +1,25 @@
...
+
+@pytest.mark.parametrize(
+    "density",
+    [
+        GaussianDensity(atomic_gaussian_width=1.2),
+        LodeDensity(atomic_gaussian_width=1.2, potential_exponent=1),
+    ],
+)
+def test_derivative(density):
+    # Exclude very small values due instabilitues of LodeDensity
```
PicoCentauri, line 19: @kvhuguenin. This might be something we have to look into. I thought we avoid these instabilities by the expansion we do for small values but it seems not super stable...

python/rascaline/tests/utils/splines.py
```diff
@@ -326,3 +326,48 @@ def test_center_contribution_gto_gaussian():
...
+        density=mydensity(atomic_gaussian_width),
+        **spliner_args,
+    )
+
+    splines_analytic = spliner_analytical.compute(n_spline_points)
+    splines_numerical = spliner_numerical.compute(n_spline_points)
+
+    n_points = len(splines_analytic["TabulatedRadialIntegral"]["points"])
+
+    # TODO fix 0 component!!
```
PicoCentauri, line 365: @kvhuguenin also this puzzles me a bit. Everything is fine, but the spline `derivatives` for r=0 are not very accurate for some values.

python/rascaline/tests/utils/atomic_density.py
PicoCentauri: This was only a warning issued by `np.where`. I filter these warnings now.

python/rascaline/examples/long-range-descriptor.py
```diff
@@ -0,0 +1,214 @@
...
+max_radial = 5
+max_angular = 1
+atomic_gaussian_width = 1.2
+recalc_splines = True
+center_atom_weight = 1.0
+
+
+# %%
+# We choose a relatively low spline accuracy (default is ``1e-8``) to achieve a quick
+# compuyattion of the spline points. You can increase spline accuracy if required,
```
SanggyuChong, line 69: typo

SanggyuChong (commented):
the notebook looks good to me. imo it would be nicer if we also show a case where an atom exists outside of the cutoff, but the current version does a good job at showing that what needs to happen indeed takes place.

python/rascaline/examples/long-range-descriptor.py
PicoCentauri: Thanks!

SanggyuChong:
Thanks again for this wonderful PR. Everything is heading towards the right direction, but I'm afraid the accuracy is still not where it should be. Can you shed some light on where the accuracy bottleneck is? I've played around with `spline_accuracy` and `k_cutoff` but have seen no improvement in terms of the "residue" between real space and fourier space for the monomer case.

SanggyuChong:
So -- never mind what I said up there, further experimentation and a quick chat with @Luthaf have revealed that the residual values could mostly be coming from the periodic images of the monomer. Still looking into this, but can confirm already that this is the primary source of the residue.

I suggest that @PicoCentauri can make a mention of this in the How-to guide, and then with the approving reviews, this gets merged in :+1:

docs/src/how-to/long-range.rst
```diff
@@ -0,0 +1,45 @@
+.. _userdoc-how-to-long-range-descriptor:
+
+Pure LODE long-range descriptor
```
Luthaf, line 3: I'm not sure if "pure LODE" is the best way to call this descriptor. Maybe "long-range only LODE descriptor"?

docs/src/how-to/long-range.rst
```diff
@@ -0,0 +1,45 @@
...
+    :align: center
+    :class: only-light
+
+.. figure:: ../../static/images/long-range-descriptor-dark.*
+    :align: center
+    :class: only-dark
+
+This descriptor can be particular useful when already having a good descriptor for the
+short-range density like (SOAP) and the long-range descriptor (far field) should not
+have the information the short-range descriptor already contains.
```
Luthaf, line 24:
```suggestion
short-range density like (SOAP) and the long-range descriptor (far field) should contain
different information from what the short-range descriptor already offers. 
```

docs/static/images/long-range-descriptor-dark.pdf
Luthaf, line 1: this is a bit big to include in the repo.

docs/static/images/long-range-descriptor-dark.svg
Luthaf, line 1: seems that this does not contain the full svg file?

python/rascaline/examples/long-range-descriptor.py
```diff
@@ -0,0 +1,365 @@
...
+
+
+spline_accuracy = 1e-2
+
+
+# %%
+# As a projection basis, we don't use the usual :py:class:`GtoBasis
+# <rascaline.utils.GtoBasis>` which is commonly used for short range descriptors.
+# Instead, we select the :py:class:`MonomialBasis <rascaline.utils.MonomialBasis>` which
+# is the optimal basis for the LODE descriptor.
```
Luthaf, line 100: Why is it the optimal basis for LODE? Could you maybe add a link to the corresponding proof/paper?

python/rascaline/examples/long-range-descriptor.py
```diff
@@ -0,0 +1,365 @@
...
+
+
+basis = MonomialBasis(cutoff=cutoff)
+
+
+# %%
+# For the density, we choose a smeared power law as used in LODE, which does not decay
+# exponentially like a :py:class:`Gaussian density <rascaline.utils.GaussianDensity>`
+# and is therefore suited to describe long-range interactions between atoms.
+
```
Luthaf, line 110: might be nice to plot this density & the Gaussian density to make this point clearer

python/rascaline/examples/long-range-descriptor.py
```diff
@@ -0,0 +1,365 @@
...
+    cutoff=cutoff,
+    max_radial=max_radial,
+    max_angular=max_angular,
+    atomic_gaussian_width=atomic_gaussian_width,
+    radial_basis=rs_splines,
+    center_atom_weight=center_atom_weight,
+    cutoff_function={"Step": {}},
+)
+
+rs_se = rs_calculator.compute(atoms)
```
Luthaf, line 165:
```suggestion
real_space_expansion = rs_calculator.compute(atoms)
```

python/rascaline/examples/long-range-descriptor.py
```diff
@@ -0,0 +1,365 @@
...
+
+# %%
+# You might want to save the spline points using :py:func:`json.dump` to a file to reuse
+# them later without recalculating them. Saving them is especially useful if the spline
+# calculations are expensive, i.e., if you increase the ``spline_accuracy``.
+#
+# With the spline points on hand, we now compute the real space spherical expansion
+
+
+rs_calculator = SphericalExpansion(
```
Luthaf, line 155:
```suggestion
real_space_calculator = SphericalExpansion(
```

python/rascaline/examples/long-range-descriptor.py
```diff
@@ -0,0 +1,365 @@
...
+)
+
+rs_se = rs_calculator.compute(atoms)
+
+
+# %%
+# and the Fourier Space / LODE Spherical expansion.
+
+
+fs_calculator = LodeSphericalExpansion(
```
Luthaf, line 172:
```suggestion
fourier_space_calculator = LodeSphericalExpansion(
```

python/rascaline/examples/long-range-descriptor.py
```diff
@@ -0,0 +1,365 @@
...
+    max_radial=max_radial,
+    max_angular=max_angular,
+    atomic_gaussian_width=atomic_gaussian_width,
+    center_atom_weight=center_atom_weight,
+    potential_exponent=potential_exponent,
+    radial_basis=fs_splines,
+    k_cutoff=k_cutoff,
+)
+
+fs_se = fs_calculator.compute(atoms)
```
Luthaf, line 183:
```suggestion
fourier_space_expansion = fs_calculator.compute(atoms)
```

PicoCentauri:
> I suggest that @PicoCentauri can make a mention of this in the How-to guide

Yes, I can add a note that any residual contributions might originate from the periodic images. To verify this you can either scale the box or increase the potential exponent.

> Can you shed some light on where the accuracy bottleneck is?

Besides the periodicity: Another thing for the accuracy, we haven't explored at all, are the accuracy options in [quad](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.quad.html) (`epsabs` and `epsrel`) itself. The default values are already quite small (`1e-8`) but we can keep this in mind if accuracy problems appear for splining again.

docs/static/images/long-range-descriptor-dark.pdf
PicoCentauri: Let me do a real vector version. This is very ugly version that uses the pixel figures from the paper.

python/rascaline/examples/long-range-descriptor.py
PicoCentauri: Good point. I added this to the new version.

python/rascaline/examples/long-range-descriptor.py
PicoCentauri: Yep! A log-log plot should now be in.

docs/src/how-to/long-range.rst
PicoCentauri: Yes, I agree.

python/rascaline/examples/long-range-descriptor.py
```diff
@@ -0,0 +1,424 @@
...
+
+
+# %%
+# The plot shows that the spherical expansion for the Fourier space is non-zero while
+# the difference between the two expansions is very small.
+#
+# .. warning::
+#   Small residual values may stems from the contribution of the periodic images. You
+#   can verify and reduce those contributions by either increasing the cell and/or
+#   increase the ``potential_exponent``.
```
PicoCentauri, line 305: @SanggyuChong are you happy with this "warning"?

docs/static/images/long-range-descriptor-dark.pdf
Luthaf, line 1: could we have a single version of this file by making the text "medium" grey? Something readable on both light & dark theme.

python/rascaline/examples/long-range-descriptor.py
```diff
@@ -0,0 +1,427 @@
...
+    ax[0].set_title(f"center_type={key.values[0]}\n Fourier space sph. expansion")
+    im = ax[0].matshow(values_fourier_space, vmin=vmin, vmax=vmax)
+    ax[0].set_ylabel("sample index")
+
+    ax[1].set_title("Difference between Fourier and real space sph. expansion")
+    ax[1].matshow(values_subtracted, vmin=vmin, vmax=vmax)
+    ax[1].set_ylabel("sample index")
+    ax[1].set_xlabel("property index")
+
+    fig.colorbar(im, ax=ax[0], orientation="horizontal", fraction=0.1, label="values")
```
Luthaf, line 420: this plot looks weird in the rendered documentation, the left & right side have different width. Is this intentional?

python/rascaline/examples/long-range-descriptor.py
```diff
@@ -0,0 +1,427 @@
...
+
+# %%
+# We clearly see that the values of the subtracted spherical are much larger compared to
```
Luthaf, line 423: the color difference is not that large, since the color bar is capped by very high values for low n. Maybe we could clamp the values between -0.25 & 0.5 to add more color distinctions for the low values?

docs/static/images/long-range-descriptor-dark.pdf
PicoCentauri: Sure.

python/rascaline/examples/long-range-descriptor.py
PicoCentauri: No, idea what sphinx-gallery and matplotlib are doing here. I tried resizing the figures but it didn't help. I am now showing them in different "cells".

Luthaf (approved):

PR #301, opened 2024-04-22
Untangle strain and cell gradients

Luthaf:
This partially undo #105: we where right to use gradient w.r.t. strain there, but these should be stored using `"strain"` as the gradient parameter.

Now, we can have three kinds of gradients:

- w.r.t. `"positions"`, taken at fixed cell matrix;
- w.r.t. `"cell"`, taken at fixed Cartesian positions;
- w.r.t. `"strain"`, following the definition for the virial.

The first two are then used when integrating with Torch's autograd system; and the last one can be useful when computing gradients in forward mode.

As a side effect, I fixed gradients of SOAP when the cell is smaller than the cutoff, we where doing some summations wrong in this case.


python/rascaline/rascaline/calculator_base.py
```diff
@@ -197,100 +197,104 @@ def compute(
...
-            an empty list ``[]``, no gradients are computed. Gradients are
-            stored inside the different blocks, and can be accessed with
-            ``descriptor.block(...).gradient(<parameter>)``, where
-            ``<parameter>`` is ``"positions"`` or ``"cell"``. The following
-            gradients are available:
+        :param gradients: List of gradients to compute. If this is :py:obj:`None` or an
+            empty list ``[]``, no gradients are computed. Gradients are stored inside
+            the different blocks, and can be accessed with
+            ``descriptor.block(...).gradient(<parameter>)``, where ``<parameter>`` is
+            ``"positions"``, ``"cell"`` or ``"strain"``. The following gradients are
```
PicoCentauri, line 214: Since the gradients user want to calculate for virials are gradients wrt to strain we should maybe swap the order of mentioning to: 

```suggestion
            ``"positions"``, ``"strain"`` or ``"cell"``. The following gradients are
```

this also reflects how to list them below.

rascaline/src/calculators/mod.rs
```diff
@@ -32,7 +32,8 @@ pub trait CalculatorBase: std::panic::RefUnwindSafe {
     fn samples(&self, keys: &Labels, systems: &mut [Box<dyn System>]) -> Result<Vec<Labels>, Error>;
 
     /// Can this calculator compute gradients with respect to the `parameter`?
-    /// Right now, `parameter` can be either `"positions"` or `"cell"`.
+    /// Right now, `parameter` can be either `"positions"`, `"cell"` or
+    /// `"strain"`.
```
PicoCentauri, line 36:
```suggestion
    /// Right now, `parameter` can be either `"positions"`, `"strain"` or
    /// `"cell"`.
```

rascaline/src/calculators/soap/power_spectrum.rs
```diff
@@ -677,73 +681,90 @@ impl CalculatorBase for SoapPowerSpectrum {
                     });
             }
 
-            // gradients with respect to the cell parameters
-            if let Some(mut gradient) = block.gradient_mut("cell") {
-                let gradient = gradient.data_mut();
+            // gradients with respect to the strain/cell
+            for parameter in ["cell", "strain"] {
```
PicoCentauri, line 685: Shouldn't the list `["cell", "strain"]` be created from what the user is requesting? If he only want `"strain"` we should not do cell gradients

rascaline/src/calculators/soap/radial_spectrum.rs
```diff
@@ -165,7 +165,7 @@ impl CalculatorBase for SoapRadialSpectrum {
 
     fn supports_gradient(&self, parameter: &str) -> bool {
         match parameter {
-            "positions" | "cell" => true,
+            "positions" | "cell" | "strain" => true,
```
PicoCentauri, line 168:
```suggestion
            "positions" | "strain" | "cell"  => true,
```

rascaline/src/calculators/atomic_composition.rs
```diff
@@ -75,8 +75,7 @@ impl CalculatorBase for AtomicComposition {
 
     fn supports_gradient(&self, parameter: &str) -> bool {
         match parameter {
-            "positions" => true,
-            "cell" => true,
+            "positions" | "cell" | "strain" => true,
```
PicoCentauri, line 78:
```suggestion
            "positions" | "strain" | "cell" => true,
```

rascaline/src/calculators/soap/spherical_expansion.rs
```diff
@@ -198,38 +188,60 @@ impl SphericalExpansion {
...
                         let mut cell_gradients = cell_gradients.slice_mut(
                             s![neighbor_type_i, mapped_center, .., .., .., ..]
                         );
 
-                        for spatial_1 in 0..3 {
-                            for spatial_2 in 0..3 {
-                                let inverse_cell_pair_vector_2 = inverse_cell_pair_vector[spatial_2];
+                        for abc in 0..3 {
+                            let shift = pair.cell_shift_indices[abc] as f64;
+                            for xyz in 0..3 {
```
PicoCentauri, line 210: I think before you are using `xyz_1` and `xyz_2` and here seems like `abc` and `xyz`. Maybe we should be consistent.

rascaline/src/calculators/soap/spherical_expansion.rs
```diff
@@ -262,30 +273,55 @@ impl SphericalExpansion {
...
```
PicoCentauri, line 290: same as above

rascaline/src/calculators/soap/spherical_expansion.rs
```diff
@@ -639,8 +693,7 @@ impl CalculatorBase for SphericalExpansion {
 
     fn supports_gradient(&self, parameter: &str) -> bool {
         match parameter {
-            "positions" => true,
-            "cell" => true,
+            "positions" | "cell" | "strain" => true,
```
PicoCentauri, line 696:
```suggestion
            "positions" | "strain" | "cell" => true,
```

rascaline/src/calculators/tests_utils.rs
```diff
@@ -328,46 +329,106 @@ pub fn finite_differences_cell(mut calculator: Calculator, system: &SimpleSystem
     };
     let reference = calculator.compute(&mut [Box::new(system.clone())], calculation_options).unwrap();
     let original_cell = system.cell().unwrap().matrix();
-    let original_cell_inverse = original_cell.inverse();
 
-    for spatial_1 in 0..3 {
-        for spatial_2 in 0..3 {
+    for abc in 0..3 {
+        for xyz in 0..3 {
```
PicoCentauri, line 334: maybe again...

```suggestion
    for xyz_1 in 0..3 {
        for xyz_2 in 0..3 {
```

python/rascaline/rascaline/calculator_base.py
```diff
@@ -197,100 +197,104 @@ def compute(
...
+                  \frac{\partial \langle q \vert A_i \rangle}
+                       {\partial \mathbf{\epsilon}}
 
-        :param selected_samples: Set of samples on which to run the calculation.
-            Use :py:obj:`None` to run the calculation on all samples in the
-            ``systems`` (this is the default).
+            - ``"cell"``, for gradients of the representation with respect to the
+              system's cell parameters. These gradients are computed at fixed positions,
+              and often not what you want when computing gradients explicitly (they are
+              mainly used in ``rascaline.torch`` to integrate with backward
```
PicoCentauri, line 253: Should be maybe mention here again that for stresses one should use gradients with respect to `strain`. I know that this is listed above but just to be sure.

PicoCentauri (commented):
Very nice!

rascaline/src/calculators/soap/power_spectrum.rs
Luthaf: This will be checked by the `if let Some(mut gradient) = block.gradient_mut(parameter)`. The block contains the gradients the user requested; and this function returns `None` if no gradient is defined.

rascaline/src/calculators/soap/spherical_expansion.rs
Luthaf: this is intentional: the components are called `xyz_1` and `xyz_2` for strain gradients, but `abc` and `xyz` for the cell gradients.

rascaline/src/calculators/soap/power_spectrum.rs
PicoCentauri: Okay makes sense.

rascaline/src/calculators/soap/spherical_expansion.rs
Luthaf: Not yes, that's something I need to add. They match https://lab-cosmo.github.io/metatensor/latest/atomistic/outputs.html, but should also be documented here.

rascaline/src/calculators/soap/radial_spectrum.rs
Luthaf: this is fully internal, I don't think the order matters

PicoCentauri: true it was just my consistency me that wanted to change this.

rascaline/src/calculators/soap/spherical_expansion.rs
PicoCentauri: Okay are we doing this here or in another PR?

Luthaf: I'll open an issue and do this in a separate PR =)

bananenpampe:
Hello, is it possible that these changes are breaking changes with respect to older rascaline versions?
I am seeing broken NPT simulations with the newest rascaline version

bananenpampe:
Old rascaline:
Bildschirmfoto 2024-05-30 um 01 29 28

New rascaline:
Bildschirmfoto 2024-05-30 um 01 30 39

unwrapped

Luthaf:
which simulation engine/driver are you using? The old i-PI driver used for the alchemical HEA project was cancelling the issue fixed here and producing the right gradients overall. But if you use it with the code from this PR then the correction will make it wrong.

bananenpampe:
Okay, is it off by only a scaling factor with the cell-volume, or is it more that needs to be changed ?

Luthaf:
Two things needs to change: the code needs to compute `"strain"` gradients instead of `"cell"` gradients (the old cell gradients where almost the same as the new strain gradients), and then scale by the volume.

PR #306, opened 2024-05-29
Update installation.rst

bananenpampe:
Adds updated rascaline-torch installation instructions

----
📚 Documentation preview 📚: https://rascaline--306.org.readthedocs.build/en/306/

### DIFF
--- a/docs/src/get-started/installation.rst
+++ b/docs/src/get-started/installation.rst
@@ -124,7 +124,7 @@ Building from source:
     pip install --upgrade pip
 
     # alternatively, the same thing in a single command
-    pip install git+https://github.com/luthaf/rascaline#subdirectory=python/rascaline-torch
+    pip install --extra-index-url https://download.pytorch.org/whl/cpu git+https://github.com/luthaf/rascaline#subdirectory=python/rascaline-torch
 
 
 For usage from C++


docs/src/get-started/installation.rst
```diff
@@ -124,7 +124,7 @@ Building from source:
     pip install --upgrade pip
 
     # alternatively, the same thing in a single command
-    pip install git+https://github.com/luthaf/rascaline#subdirectory=python/rascaline-torch
+    pip install --extra-index-url https://download.pytorch.org/whl/cpu git+https://github.com/luthaf/rascaline#subdirectory=python/rascaline-torch
```
Luthaf, line 127: This is not always necessary, and some people might want to install the cuda version of torch. A better solution here would be to do something like the metatensor docs: https://docs.metatensor.org/latest/installation.html#install-torch (the paragraph starting with "By default, PyTorch is installed with CUDA …")

bananenpampe: okay, sure, but as is this recommended install will fail on all machines that do not have CUDA installed which is usually the case

Luthaf: I mean, people should be able to read 4 lines of text before blindly copy/pasting an installation command =). This will also be less of an issue when we do a full release on PyPI.

Luthaf:
This should not longer be needed, now that we have wheels on PyPI

PR #308, opened 2024-05-31
temptative fix for the case where a calculator does not return anything

liam-o-marsh:
simple python test:
```python
import pyscf, rascaline
mol = rascaline.systems.wrap_system(pyscf.M(atom='H 0 0 -10; H 0 0 10'))
rascaline.calculators.NeighborList(5.0,False).compute(mol)
```

Since the cutoff is smaller than the distance between hydrogens, there are no pairs found by `compute_neighbors`, causing `assert_ne!(keys.count(), 0);` to panic (at `rascaline/src/calculator.rs:59:9`).

Those three lines prevent the panic by making `Calculator::prepare()` return an empty TensorMap early.

----
📚 Documentation preview 📚: https://rascaline--308.org.readthedocs.build/en/308/

### DIFF
--- a/rascaline/src/calculator.rs
+++ b/rascaline/src/calculator.rs
@@ -294,6 +294,10 @@ impl Calculator {
     #[time_graph::instrument(name="Calculator::prepare")]
     fn prepare(&mut self, systems: &mut [Box<dyn System>], options: CalculationOptions) -> Result<TensorMap, Error> {
         let default_keys = self.implementation.keys(systems)?;
+        if default_keys.count() == 0 {
+            return Ok(TensorMap::new(default_keys, vec![])?)
+        }
+
         let keys = match options.selected_keys {
             Some(keys) if keys.is_empty() => {
                 return Err(Error::InvalidParameter("selected keys can not be empty".into()));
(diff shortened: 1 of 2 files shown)


Luthaf:
Thanks you for the PR! This should be added as a test somewhere, either in Python or directly in the Rust code.

I'm not sure if an early return is the best solution here, since it might interact weirdly with key selections or samples/properties selection when giving predefined values. It think that returning an empty vec instead of the failing assert if there are no keys would be a nicer solution. I'll think of a test for this (slightly convoluted) case

liam-o-marsh:
https://github.com/Luthaf/rascaline/actions/runs/9319582029/job/25654465095?pr=308
the fact that it failed like this on only one platform tastes like cosmic rays?
anywho, yeah that way makes more sense

Luthaf:
The CI error does seems unrelated, not sure why it is failing though. @PicoCentauri did you seen anything like this while writing the spliner code?

Luthaf:
Ok, CI error should be fixed now. Can you rebase this branch on top of master?

Luthaf (approved):
Thanks a lot!

liam-o-marsh:
you're welcome!

PR #316, opened 2024-07-03
CGs: introduce general `TensorCorrelator` and refactor `DensityCorrelations`

jwa7:
Some changes:

1. Add a `keep_l_in_keys` parameter for the `DensityCorrelations` calculator. This defaults to false, maintaining the previous behaviour. This makes it easier to write the pair correlator as `l_{x}` and `k_{x}` combination info needs to be kept in the keys for further CG tensor products. While this adds complexity to `DensityCorrelations`, the idea is for this complexity to in the end be wrapped by higher-level, more specific calculators (i.e. a lambda-SOAP calculator), such that the normal user doesn't need to worry about complex metadata things like this.
2. `body_order` replaces `correlation_order` as the parameter in `DensityCorrelations`. `"body_order"` as a key dimension is returned in the output TensorMaps, as this is also useful to know for a general CG tensor product. Since we don't yet have global metadata, keeping it in keys is best for now.
3. A `match_keys` argument has also been added, and the core routines that precompute tensor products of keys metadata have been modified. Internally, the keys metadata dimensions are categorised into 'standard' (i.e. order_nu, o3_lambda, o3_sigma), 'CG combination' (i..e the `l` and `k` lists), and 'other' key names. Standard and combination dimensions are handled as usual, but the other dimensions are either matched (based on ``match_keys`` argument) or fully multiplied.
4. Arbitrary properties dimensions are now handled. Previously, the function assumes that (and handles by hard-coding) only "neighbor_type" and "n" properties dimensions are present in the TensorBlocks being combined. Now, arbitrary properties can be combined. The requirement is that the properties dimension names are different between the two blocks being combined. Property renaming is then handled by the wrapping method `_correlate_density` of `DensityCorrelations`. Properties of the arbitrary body order tensor are named (i.e. for radial channels) "n_1", "n_2", ..., "n_{nu - 1}", and the nu=1 tensor being combined on the current iteration is renamed so that its property dimension is "n_{nu}". Suffixing with "_{x}" is chosen and applied generically, except for dimensions suffixed with "_type", i.e. "neighbor_type" becomes "neighbor_{x}_type".

----
📚 Documentation preview 📚: https://rascaline--316.org.readthedocs.build/en/316/


python/rascaline/rascaline/utils/clebsch_gordan/_correlate_density.py
```diff
@@ -73,21 +73,34 @@ class DensityCorrelations(TorchModule):
...
+        that were combined together be kept in the keys or moved to the properties?
+
+        Keys named ``l_{i}`` correspond to the input ``components``, with ``l_1`` being
+        the last entry in ``components`` and ``l_N`` the first one. Keys named ``k_{i}``
+        correspond to intermediary spherical components created during the calculation,
+        i.e. a ``k_{i}`` used to be ``o3_lambda``.
+
+        This defaults to false for TensorMaps output at each requested iteration, such
+        that the keys are moved to properties. If wanting to use the output of this
+        class for further CG tensor products, this should be set to True.
```
Luthaf, line 95:
```suggestion
        This defaults to ``False`` for :py:class:`TensorMap` output at each requested iteration, such
        that the keys are moved to properties. If you want to use the output of this
        class for further CG tensor products, this should be set to ``True``.
```

python/rascaline/rascaline/utils/clebsch_gordan/_correlate_density.py
```diff
@@ -178,11 +192,12 @@ def __init__(
...
         # Parse the bool flags that control skipping of redundant CG combinations
         # and TensorMap output from each iteration
-        self._skip_redundant, self._output_selection = (
+        self._skip_redundant, self._output_selection, self._keep_l_in_keys = (
             _utils.parse_bool_iteration_filters(
                 n_iterations,
                 skip_redundant=skip_redundant,
                 output_selection=output_selection,
+                keep_l_in_keys=keep_l_in_keys,
             )
```
Luthaf, line 201: why are we parsing multiple thing in a single function call? Instead of 
```suggestion
        self._skip_redundant = _utils.parse_bool_iteration_filters(n_iterations, skip_redundant)
        self._output_selection = _utils.parse_bool_iteration_filters(n_iterations, output_selection)
        self._keep_l_in_keys = _utils.parse_bool_iteration_filters(n_iterations, keep_l_in_keys)
```

python/rascaline/tests/utils/correlate_density.py
```diff
@@ -261,6 +261,7 @@ def test_lambda_soap_vs_powerspectrum():
     lsoap = calculator.compute(density)
     keys = lsoap.keys.remove(name="o3_lambda")
     keys = keys.remove("o3_sigma")
+    keys = keys.remove("order_nu")
```
Luthaf, line 264: should we rename this to `body_order` instead?

jwa7: We could do, but would require a change everywhere. Different PR?

python/rascaline/rascaline/utils/clebsch_gordan/_correlate_density.py
jwa7: Main reason is because they (and their defaults) interact in some way

python/rascaline/tests/utils/correlate_density.py
jwa7: And perhaps something to be discussed with others too, "correlation_order" / "order_nu" feels more CG-like, but "body_order" more intuitive / better UX. Something that can be easily changed later but I think not necessary now. Thoughts?

python/rascaline/rascaline/utils/clebsch_gordan/_correlate_density.py
Luthaf: It seems that they don't interact actually. `parse_bool_iteration_filters` is just handling the three of them one after the other.

python/rascaline/tests/utils/correlate_density.py
Luthaf: I don't know, this feels trivial enough to change now, except if we don't agree on the new naming?

Luthaf: Since this was internal before, the name did not matter too much. But if we are exposing it, we should give it a nice name!

jwa7: Fair enough, I'll change it

python/rascaline/rascaline/utils/clebsch_gordan/_correlate_density.py
jwa7: Sure, done

python/rascaline/rascaline/utils/clebsch_gordan/_utils.py
```diff
@@ -144,62 +143,28 @@ def parse_selected_keys(
...
-        raise TypeError("`output_selection` must be passed as a `list` of `bool`")
-    if not all([isinstance(v, bool) for v in output_selection]):
-        raise TypeError("`output_selection` must be passed as a `list` of `bool`")
-
-    if isinstance(keep_l_in_keys, bool):
-        keep_l_in_keys_ = [
-            keep_l_in_keys if output else False for output in output_selection
-        ]
+    if bool_filter is None:
+        bool_filter = [False] * (n_iterations - 1) + [True]
```
Luthaf, line 153: this is only intended for output selection, right? I would then handle it outside of this function for better clarity

jwa7: Yes true

jwa7: done!

Luthaf:
Should this be closed?

PR #317, opened 2024-07-04
Implementation of `EquivariantPowerSpectrum` and `EquivariantPowerSpectrumByPair` convenience calculators

jwa7:
Implements:

1. a lambda-SOAP (i.e. `EquivariantPowerSpectrum`) calculator that wraps `DensityCorrelations`
2. a `CorrelateTensorWithDensity` calculator that performs a single CG tensor product between a tensor of arbitrary body order and a density (body-order 2) tensor, where the samples between the tensor and density are different.
3. a `EquivariantPowerSpectrumByPair` calculator, which generates a `SphericalExpansion` and `SphericalExpansionByPair`, manipulates the metadata appropriately, then uses `CorrelateTensorWithDensity` calculator to perform `rho_i x g_ij` to generate an Equivariant power spectrum by pair

----
📚 Documentation preview 📚: https://rascaline--317.org.readthedocs.build/en/317/

### DIFF
--- a/python/rascaline/rascaline/utils/clebsch_gordan/_lambda_soap.py
+++ b/python/rascaline/rascaline/utils/clebsch_gordan/_lambda_soap.py
@@ -8,11 +8,10 @@ from typing import List, Optional, Union
 
 import numpy as np
 
+from ...calculators import SphericalExpansion, SphericalExpansionByPair
 from .. import _dispatch
 from .._backend import (
     Labels,
-    SphericalExpansion,
-    SphericalExpansionByPair,
     TensorBlock,
     TensorMap,
     TorchModule,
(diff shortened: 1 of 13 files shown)


python/rascaline/rascaline/utils/clebsch_gordan/_lambda_soap.py
```diff
@@ -0,0 +1,288 @@
...
+    HAS_TORCH = True
+except ImportError:
+    HAS_TORCH = False
+
+# ======================================================================
+# ===== Public API functions
+# ======================================================================
+
+
+class LambdaSoap(TorchModule):
```
Luthaf, line 41: I think I would prefer something like `LambdaPowerSpectrum`/`LambdaSoapPowerSpectrum`/`EquivariantPowerSpectrum`/`EquivariantSoapPowerSpectrum`

I'm fighting against the SOAP == PowerSpectrum thinking, and making this explicit would help IMO

python/rascaline/rascaline/utils/clebsch_gordan/_lambda_soap.py
```diff
@@ -0,0 +1,288 @@
...
+    """
+    Computes a Lambda-SOAP equivariant descriptor.
+    """
+
+    def __init__(
+        self,
+        spherical_expansion_hypers: dict,
+        atom_types: Optional[List[int]] = None,
+        angular_cutoff: Optional[int] = None,
```
Luthaf, line 50: do we really want this as a separate hyper for angular cutoff? Should we use the one from spx hypers instead?

python/rascaline/rascaline/utils/clebsch_gordan/_lambda_soap.py
```diff
@@ -0,0 +1,288 @@
...
+        selected_keys: Optional[Labels] = None,
```
Luthaf, line 51: this should be part of `compute/forward`: different calls to the same calculator might request different keys

python/rascaline/rascaline/utils/clebsch_gordan/_lambda_soap.py
```diff
@@ -0,0 +1,288 @@
...
+            )
+        density = density.keys_to_properties(keys_to_move)
+
+        # Compute the nu = 2 lambda-SOAP descriptor
+        lsoap = self._density_correlations_calc.compute(density)
+
+        return lsoap
+
+
+class LambdaSoapXSphExByPair(TorchModule):
```
Luthaf, line 111: This is `\rho^2 x g_ij`, right? since `g_ij` is `SphericalExpansionByPair`, this should be `BispectrumByPair` IMO (with potential `Lambda`, `Soap`, `Equivariant` in there as well).

Luthaf (commented):
I only looked at naming! One other thing that we could do here would be `SoapBiSpectrum`, for the invariant only-part

PR #327, opened 2024-09-19
Use Labels.select for key selection in ClebschGordanProduct

Luthaf:
This is a follow-up to #316 

----
📚 Documentation preview 📚: https://rascaline--327.org.readthedocs.build/en/327/


jwa7 (approved):
Much cleaner!

PR #337, opened 2024-10-30
systems: fix pyscf system wrapper in case pbc not imported

liam-o-marsh:
should really caught this months ago


liam-o-marsh:
wait.
now that I think about it more

liam-o-marsh:
there. less shoddy fix.

Luthaf (approved):
Looks good, thanks!

PR #328, opened 2024-09-19
Update hyper-parameters to a more extensible schema

Luthaf:
There are two main ingredients to a spherical expansion:

- the density being expanded
- the basis on which to make the expansion

Additionally, real space/short range (i.e. SOAP) density expansion requires the definition of local environments with a cutoff radius and smoothing functions.

This PR refactor the hyper-parameters of spherical expansion and related calculators to reflect this structure and make is easier to add new variations in the density and basis ingredients.

TBD:

- [x] code to automatically convert from old hypers
- [x] update examples & tutorials
- [x] rewrite all of splining code
- [x] should we expose classes for the hyper building blocks? I.e. allow something like 

```py
cutoff = Cutoff(
    radius=5.0,
    smoothing=ShiftedCosineSmoothing(width=0.5),
)

density = GaussianDensity(width=0.3)

basis = TensorProductBasis(
    max_angular=4,
    radial=GtoRadialBasis(max_radial=6),
)

calculator = SoapPowerSpectrum(cutoff, density, basis)
```

Issue #284, opened 2024-02-23
Path to: `rascaline` release


jwa7:
Here we outline the steps needed to get to a rascaline release.

- [ ] Build steps, outlined in #281 
- [ ] Renaming metadata to match new conventions in metatensor
- [ ] Find a product name, free on PyPI, GitHub, [crate.io](https://crates.io/), no problematic related Google results

Luthaf:
- [ ] Separate `cell` and `strain` gradients and fix cell gradients for small cell/large cutoff calculations

jwa7:
- [ ] Agree on `k_{x}` numbering convention between the `cartesian_to_spherical` routine(s) and CG utils
- [ ] Generalised and re-factored CG tensor product routines, to be done in #316 
- [ ] Implement high level CG-based calculators for the main use cases, fully documented and tested

Luthaf:
I don't think that anything related to the CG routines is related to the release. This issue is about stuff we can NOT do a release without, but CG can always be added/modified after an initial release (that's what incompatible version numbers are for). If you want to keep track of everything CG related somewhere, could you open a separate issue?

PR #339, opened 2024-10-31
rebasing system-with-data onto main

liam-o-marsh:
----
📚 Documentation preview 📚: https://rascaline--339.org.readthedocs.build/en/339/

### DIFF
--- a/rascaline-c-api/src/calculator.rs
+++ b/rascaline-c-api/src/calculator.rs
@@ -402,7 +402,7 @@ pub unsafe extern fn rascal_calculator_compute(
         }
         check_pointers!(calculator, descriptor, systems);
 
-        // Create a Vec<Box<dyn System>> from the passed systems
+        // Create a Vec<System> from the passed systems
         let c_systems = if systems_count == 0 {
             &mut []
         } else {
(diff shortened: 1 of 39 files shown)


liam-o-marsh:
honestly having this be a merge proper is a bad idea, since my version is already a rebase
it is probably better to resolve this out-of-band, if you have the time for it
or not at all, if you don't want to merge system-with-data into main without my bond-centered stuff to use it

Luthaf:
I can do a quick rebase of this branch! For the bond-centered descriptor, I am finally able to spend some time working on rascaline, so after #328 is merged my next priority will be making a proper release of the code (#284). At the same time/after this, we could try again to update the bond-centered descriptor to the new hyper schema & merge it. 

Since the last time we chatted about it, was this descriptor used in a scientific paper? It would push it a couple of steps up in my TODO list!

Luthaf:
Ok, the branch has been rebased, and also includes #338.

liam-o-marsh:
thanks!

> At the same time/after this, we could try again to update the bond-centered descriptor to the new hyper schema & merge it.  
> Since the last time we chatted about it, was this descriptor used in a scientific paper? It would push it a couple of steps up in my TODO list!

Definitely later, that project became too big to be a single publication, and somehow actual bond-centered descriptions are way up in the dependency chain between papers
The main reason why I'm looking at rascaline again is because Joe told me about a bunch of changes that would be interesting, so I'm trying to slowly rebase my copy of it (which has a bunch of changes, both unmerged and not upstreamed at all)

PR #340, opened 2024-11-01
Fix inaccuracy in `rotation_adapted.rst`

frostedoyster:
If it were real, we would need m1 and m2 indices

----
📚 Documentation preview 📚: https://rascaline--340.org.readthedocs.build/en/340/


jwa7 (approved):
Thanks! (and thanks Michelangelo)

Issue #281, opened 2024-02-12
Port over build system improvements from metatensor


Luthaf:
There have been a bunch of build system and CI improvements in metatensor that we should port to rascaline as well

- https://github.com/lab-cosmo/metatensor/pull/382
- https://github.com/lab-cosmo/metatensor/pull/491
- https://github.com/lab-cosmo/metatensor/pull/497
- https://github.com/lab-cosmo/metatensor/pull/506

- maybe https://github.com/lab-cosmo/metatensor/pull/406

PR #349, opened 2024-11-22
Merge featomic and featomic-c-api

Luthaf:
This will allow to create a single archive containing all the relevant sources, to be included in the sdist for the python package. 

This is part of #281

----
📚 Documentation preview 📚: https://featomic--349.org.readthedocs.build/en/349/

⚙️ [Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/metatensor/featomic/actions/artifacts/2233843849.zip)


featomic/benches/lode-spherical-expansion.rs
```diff
@@ -1,24 +1,27 @@
...
+use chemfiles::{Frame, Trajectory};
 
 use criterion::{BenchmarkGroup, Criterion, measurement::WallTime, SamplingMode};
 use criterion::{criterion_group, criterion_main};
 
-fn load_systems(path: &str) -> Vec<Box<dyn System>> {
-    let systems = featomic::systems::read_from_file(format!("benches/data/{}", path))
-        .expect("failed to read file");
+fn read_systems_from_file(path: &str) -> Vec<Box<dyn System>> {
+    let mut trajectory = Trajectory::open(path, 'r').expect("could not open the trajectory");
```
PicoCentauri, line 9: Does this really have to be mutable? You are just reading, right?

featomic/build.rs
```diff
@@ -0,0 +1,47 @@
...
+fn main() {
+    if cfg!(feature = "c-api") {
+        let crate_dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
+
+        let generated_comment = "\
+/* ============    Automatically generated file, DOT NOT EDIT.    ============ *
+ *                                                                             *
+ *    This file is automatically generated from the featomic sources,          *
+ *    using cbindgen. If you want to make change to this file (including       *
+ *    documentation), make the corresponding changes in the rust sources.      *
```
PicoCentauri, line 14: Can we give a link to the rust file here?

featomic/benches/lode-spherical-expansion.rs
Luthaf: Yes, it does because when reading a file we are modifying the internal data structure representing this file. See https://doc.rust-lang.org/std/fs/struct.File.html for example, where the `File` struct is mutable even when reading.

featomic/build.rs
Luthaf: done

PicoCentauri (approved):

PR #353, opened 2024-12-04
Fix the definition of `real2complex` for spherical harmonics

Luthaf:
We where using the complex conjugate of the Wikipedia convention. This was found thanks to @MichelangeloDomina!

Using the wrong convention means we where adding a minus sign for some of the components, which should have limited impact on the behavior of ML models, but better to correct it anyway.

----
📚 Documentation preview 📚: https://featomic--353.org.readthedocs.build/en/353/

⚙️ [Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/metatensor/featomic/actions/artifacts/2273873973.zip)


frostedoyster:
What type of model is this PR going to break, if any?

Luthaf:
So anything using the CG tensor product will need to be retrained. If you are not using something from `featomic.clebsch_gordan` (previously `rascaline.utils.clebsch_gordan`), you should be good.

jwa7 (approved):
Ran some light model training from the rholearn example and nothing seems to be different from before, so looks good! Thank you both for finding and fixing @Luthaf @MichelangeloDomina :)

PR #376, opened 2025-01-29
Adds sample selection to equivariant power spectrum

bananenpampe:
Adds sample selection to the equivariant power spectrum convenience class.

### DIFF
--- a/python/featomic/tests/clebsch_gordan/equivariant_power_spectrum.py
+++ b/python/featomic/tests/clebsch_gordan/equivariant_power_spectrum.py
@@ -163,7 +163,9 @@ def tests_sample_selection() -> None:
         frame, neighbors_to_properties=True, selected_samples=label_2nd
     )
 
-    powspec_3 = metatensor.join([powspec_1, powspec_2], axis="samples", remove_tensor_name=True)
+    powspec_3 = metatensor.join(
+        [powspec_1, powspec_2], axis="samples", remove_tensor_name=True
+    )
     powspec_4 = powspec_calc.compute(frame, neighbors_to_properties=True)
 
     assert metatensor.equal(powspec_3, powspec_4)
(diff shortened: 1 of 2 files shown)


python/featomic/tests/clebsch_gordan/equivariant_power_spectrum.py
```diff
@@ -4,6 +4,7 @@
 import numpy as np
 import pytest
 from metatensor import Labels, TensorBlock, TensorMap
+from metatensor.operations import equal, join
```
Luthaf, line 7: Please don't import from `metatensor.operations`. This could be written 

```
from metatensor import equal, join
```

Or even better

```suggestion
import metatensor as mts
```

and down the line `mts.equal`, `mts.join`

(like you would not do `from numpy import log` but instead use `np.log` where required)

Luthaf:
Thanks, looks very good overall! I just have one style comment about importing metatensor operations.

python/featomic/tests/clebsch_gordan/equivariant_power_spectrum.py
```diff
@@ -136,6 +136,41 @@ def test_equivariant_power_spectrum_neighbors_to_properties():
     metatensor.equal_raise(powspec_1, powspec_2)
 
 
+def tests_sample_selection() -> None:
```
jwa7, line 139: Just wondering if this would be run by pytest if not called `test_*` @Luthaf ?
```suggestion
def test_sample_selection() -> None:
```

jwa7 (commented):
Thanks @bananenpampe !

bananenpampe:
_test_ makes the naming consitent with the other tests

python/featomic/tests/clebsch_gordan/equivariant_power_spectrum.py
Luthaf: ah, no it will not run. `pytest` needs the function to be named `test_whatever`

Good catch!

Luthaf (approved):

PR #390, opened 2025-04-28
Tutorials for high-level calculators from the `clebsch_gordan` module

jwa7:
[📚 Download documentation for this pull-request](https://nightly.link/metatensor/featomic/actions/artifacts/3528031245.zip)


jwa7:
In the interest of not letting this stagnate, I think it's best to get the higher-level tutorials merged first and then I'll come back to the lower-level CGTensorProduct tutorial when I have time

